Issue #45, opened 2022-10-04
Deployment python script


asynchroza:
option 1:
* `docker-compose up -d`
* CRON check for updates on master, if there is one `dc down`, pull, `dc up -d`
* CRON checking for 200 http code, if down, restart `dc`

option 2:
* endpoint listening for authenticated calls from github actions on push -> VM pulls latest and restarts docker-compose
* CRON for checking 200 http code, if down, restart `dc` (we can manufacture an endpoint which pretty much does the same but returns an image which could be used as a badge on the readme (`thehub-aubg.com is up` or something like that)

asynchroza:
OPTIMIZATIONS:
* Make script track change of whether there is a currently running build

PR #46, opened 2022-10-04
#45 deployment script

asynchroza:
### DEPLOYMENT SCRIPT

This script manages the state of the build by doing requests towards localhost. If one of the services is unreachable, a rebuild is going to be triggered. 

Furthermore, the script checks for new changes within the branch on which the directory is left on and if there are any, it pulls them. Unfortunately, the `manage_app.py` script cannot be hot reloaded which would mean that further changes to the deployment script will need manual restart of the script.


manage_app.py
```diff
@@ -0,0 +1,110 @@
...
+            get_web = requests.get("http://127.0.0.1:80")
+            print(bcolors.YELLOW_IN + str(get_web) + bcolors.CEND)
+        except Exception as e:
+            
+            errors.append(e)
+            print(bcolors.RED_IN + str(e) + bcolors.CEND)
+
+
+        
+        # "connection reset by peer"
```
asynchroza, line 65: Should be investigated. Sleep timer >= 3 seconds seems to be fixing the issue.

manage_app.py
```diff
@@ -0,0 +1,110 @@
...
+            return
+
+        else:
+            # docker-compose keeps running when there is a failed container
+            stop_docker_compose()
+    
+    
+    print(bcolors.RED_IN + "BUILD FAILED" + bcolors.CEND)
+    msg['Subject'] = 'SPA BUILD FAILED'
+    msg.attach(MIMEText('<p>' + str(errors) + '</p>', 'html'))
```
asynchroza, line 101: For error in errors create a new bullet point with exception

Issue #53, opened 2022-10-10
Daily generation of a new AUTH_TOKEN in API for security reasons


asynchroza:
* Do not fetch token from .env 
* Generate a random hashed string every 24 hours and store it in memory

Code could be found in `packages/api/configs/env.go`

PR #54, opened 2022-10-10
#53 daily generation of authentication token

IvanObreshkov:
Created a function that generates a random string with a fixed length (I have used 32 characters but it can be changed by adding a different parameter to the function). Also created a cron job that sets a new random value of the BEARER TOKEN every 24 hours.

### DIFF
--- a/packages/api/configs/cron.go
+++ b/packages/api/configs/cron.go
@@ -1,0 +1,20 @@
+package configs
+
+import (
+	"fmt"
+	"time"
+
+	"github.com/go-co-op/gocron"
+)
+
+func RunCronJobs() {
+	schedule := gocron.NewScheduler(time.UTC)
+	schedule.Every(5).Seconds().Do(func() {
+		token = GenerateToken(32)
+		//this is for test
+		fmt.Println(token)
+	})
+
+	//This will block THIS Thread and allow other threads (the main thread) to continue work, right?
+	schedule.StartBlocking()
+}
--- a/packages/api/configs/cron.go
+++ b/packages/api/configs/cron.go
@@ -6,14 +6,12 @@ import (
 	"github.com/go-co-op/gocron"
 )
 
+// Sets a new random value of the BEARER TOKEN every 24 hours
 func RunCronJobs() {
 	schedule := gocron.NewScheduler(time.UTC)
-	schedule.Every(5).Seconds().Do(func() {
+    schedule.Every(1).Day().At("7:00").Do(func() {
 		SetToken()
-		//fmt.Println(token)
 	})
 
-	//This will block THIS Thread and allow other threads (the main thread) to continue work, right?
 	schedule.StartAsync()
-
 }
--- a/packages/api/configs/cron.go
+++ b/packages/api/configs/cron.go
@@ -1,7 +1,6 @@
 package configs
 
 import (
-	"fmt"
 	"time"
 
 	"github.com/go-co-op/gocron"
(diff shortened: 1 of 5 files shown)


packages/api/configs/env.go
```diff
@@ -2,25 +2,44 @@ package configs
...
-		log.Fatalf("Error loading .env file")
+var token string = GenerateToken(32)
+
+func GenerateToken(n int) string {
+	rand.Seed(time.Now().UnixNano())
+
+	random_string := "TOKEN"
+	// implement logic for generating a rand 20 char string
+
+	var chars = []rune("abcdefghijklmnopqrstuvwxyzABCDEFGHIJKLMNOPQRSTUVWXYZ0987654321")
```
asynchroza, line 20: Could you please add the following symbols `!@#$%^&*()_+-=~?><:"'{}[]|\`

packages/api/configs/env.go
```diff
@@ -2,25 +2,44 @@ package configs
...
 	"github.com/joho/godotenv"
 )
 
-func EnvMongoURI() string {
-	err := godotenv.Load()
-	if err != nil {
-		log.Fatalf("Error loading .env file")
+var token string = GenerateToken(32)
+
+func GenerateToken(n int) string {
```
asynchroza, line 14: Could you give the `n` variable a more descriptive name?

asynchroza (changes requested):

packages/api/configs/cron.go
```diff
@@ -0,0 +1,20 @@
...
+func RunCronJobs() {
+	schedule := gocron.NewScheduler(time.UTC)
+	schedule.Every(5).Seconds().Do(func() {
+		token = GenerateToken(32)
+		//this is for test
+		fmt.Println(token)
+	})
+
+	//This will block THIS Thread and allow other threads (the main thread) to continue work, right?
+	schedule.StartBlocking()
```
asynchroza, line 19: this should be `schedule.StartAsync()` otherwise it's going to block the thread. At least that's what I see from the documentation.

packages/api/configs/cron.go
```diff
@@ -0,0 +1,20 @@
...
+	"fmt"
+	"time"
+
+	"github.com/go-co-op/gocron"
+)
+
+func RunCronJobs() {
+	schedule := gocron.NewScheduler(time.UTC)
+	schedule.Every(5).Seconds().Do(func() {
+		token = GenerateToken(32)
```
asynchroza, line 13: here you should use the setter

asynchroza (changes requested):

packages/api/main.go
```diff
@@ -2,27 +2,27 @@ package main
...
-	routes.AdminRoute(app)
-	routes.EventsRoute(app)
-	routes.JobsRoute(app)
-	routes.ArticlesRoutes(app)
-	routes.JuryRoute(app)
-	routes.MentorsRoute(app)
-	routes.SponsorsRoute(app)
-	routes.PartnersRoute(app)
-	log.Fatal(app.Listen(":8000"))
+	// app := fiber.New()
```
asynchroza, line 14: WOW :DDDD

asynchroza:
```
Cron starter here -> ...
log.Fatal(app.Listen(":8000"))
```

packages/api/configs/cron.go
```diff
@@ -0,0 +1,16 @@
...
+
+import (
+	"time"
+
+	"github.com/go-co-op/gocron"
+)
+
+func RunCronJobs() {
+	schedule := gocron.NewScheduler(time.UTC)
+	schedule.Every(5).Seconds().Do(func() {
```
asynchroza, line 11: Cron to be called within a 24 hours time window

IvanObreshkov:
Created a function that generates a random string with a fixed length (I have used 32 characters but it can be changed by adding a different parameter to the function). Also created a cron job that sets a new random value of the BEARER TOKEN every 24 hours.

PR #93, opened 2022-11-28
Made https redirecting

IvanObreshkov:
Redirected the traffic to the HTTPS version of the website.

### DIFF
--- a/packages/web/nginx.conf
+++ b/packages/web/nginx.conf
@@ -1,7 +1,7 @@
 server {
   listen 80;
   server_name dev.thehub-aubg.com ;
-  return 301 https://dev.thehub-aubg.com$request_uri;
+  return 301 $host$request_uri;
 }
 server{
   listen 443 ssl;
--- a/packages/web/nginx.conf
+++ b/packages/web/nginx.conf
@@ -1,5 +1,9 @@
 server {
   listen 80;
+  server_name dev.thehub-aubg.com ;
+  return 301 https://dev.thehub-aubg.com$request_uri;
+}
+server{
   listen 443 ssl;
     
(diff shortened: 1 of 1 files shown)


packages/web/nginx.conf
```diff
@@ -1,5 +1,9 @@
 server {
   listen 80;
+  server_name dev.thehub-aubg.com ;
+  return 301 https://dev.thehub-aubg.com$request_uri;
```
asynchroza, line 4: This line is environment dependant. Please, use $host instead.

[Done right reference](https://github.com/AUBGTheHUB/hub-conf-signup/blob/master/packages/web/nginx.conf)

asynchroza (changes requested):

packages/web/nginx.conf
IvanObreshkov: Requesting review of changes

packages/web/nginx.conf
```diff
@@ -1,5 +1,9 @@
 server {
   listen 80;
+  server_name dev.thehub-aubg.com ;
+  return 301 $host$request_uri;
```
asynchroza, line 4: Missing https! This will be a circular redirect. 

Please, refer to this - as closely as possible -> [Done right reference](https://github.com/AUBGTheHUB/hub-conf-signup/blob/master/packages/web/nginx.conf)

Make sure you test your changes

asynchroza (changes requested):
Breaking changes!

packages/web/nginx.conf
IvanObreshkov: Made requested changes

asynchroza (approved):

asynchroza:
@IvanObreshkov you may merge now

Issue #92, opened 2022-11-27
HackAUBG button in hamburger menu [HIGH PRIORITY]


asynchroza:
## Brief description:
Anchor should be a button instead with rectangle form
## How to achieve it:
divs divs divs onClick

asynchroza:
@annakukova can you please add a screenshot of the hamburger menu design so that @DemonHashira could better understand what he has to do. 
You can paste it in a comment and it will get uploaded.

PR #100, opened 2022-12-01
#92 HackAUBG button in hamburger menu

DemonHashira:
Added HackAUBG button and redirection to /hackaubg.

### DIFF
--- a/packages/web/src/components/spa/Navigation/MobileNav/NavMobile.jsx
+++ b/packages/web/src/components/spa/Navigation/MobileNav/NavMobile.jsx
@@ -5,9 +5,17 @@ import './style.css';
 // import { IoIosClose } from 'react-icons/io';
 import { CgMenu } from 'react-icons/cg';
 import { MdOutlineClose } from 'react-icons/md';
-
 import { useState } from 'react';
+import {useNavigate} from 'react-router-dom';
+
+
 export const NavMobile = () => {
+    const navigate = useNavigate();
+    
+    const navigateToHackAUBG = () => {
+    navigate('/hackaubg');
+  };
+
     const [menuClass, setMenuClass] = useState('navmobile-menu not-displayed');
     const [closeButton, setCloseButton] = useState(
         'navmobile-button-close not-displayed'
@@ -86,11 +94,11 @@ export const NavMobile = () => {
                                 Jobs
                             </a>
                         </li>
-                        <li>
-                            <a href="#hackaubg" onClick={closeMenu}>
-                                HackAUBG
-                            </a>
-                        </li>
+                        <div className='navmobile-anchors-container'>
+                                <button className="navmobile-button-ham" type="button" onClick={navigateToHackAUBG} >
+                                    <p>HackAUBG</p>
+                                </button>
+                        </div>
                     </ul>
                 </div>
             </div>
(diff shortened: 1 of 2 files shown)


packages/web/src/components/spa/Navigation/MobileNav/NavMobile.jsx
```diff
@@ -86,11 +94,11 @@ export const NavMobile = () => {
...
                             </a>
                         </li>
-                        <li>
-                            <a href="#hackaubg" onClick={closeMenu}>
-                                HackAUBG
-                            </a>
-                        </li>
+                        <div className='navmobile-anchors-container'>
+                                <button className="navmobile-button-ham" type="button" onClick={navigateToHackAUBG} >
+                                    <p>HackAUBG</p>
```
asynchroza, line 99: You don't need `<p>` here.

packages/web/src/components/spa/Navigation/MobileNav/style.css
```diff
@@ -9,6 +9,31 @@
     color: white !important;
 }
 
+.navmobile-button-ham p {
```
asynchroza, line 12: This is redundant. Add the font-size style in the `.navmobile-button-ham` class.

packages/web/src/components/spa/Navigation/MobileNav/style.css
```diff
@@ -9,6 +9,31 @@
...
+    cursor: pointer;
+    margin: 20px;
+    height: 55px;
+    text-align:center;
+    border: none;
+    background-size: 300% 100%;
+    border-radius: 40px;
+}
+
+.navmobile-button-ham {
```
asynchroza, line 32: Why do we have `.navmobile-button-ham` two times? Merge these two.

packages/web/src/components/spa/Navigation/MobileNav/style.css
```diff
@@ -9,6 +9,31 @@
     color: white !important;
 }
 
+.navmobile-button-ham p {
+    padding: 0;
+    margin: 0;
+    font-size: xx-large;
+}
+
+.navmobile-button-ham {
```
asynchroza, line 18: Have you tested the responsiveness of this button? I feel like there are too many fixed values which would break on different screens.

packages/web/src/components/spa/Navigation/MobileNav/style.css
```diff
@@ -9,6 +9,31 @@
...
+    width: 200px;
+    font-size: 16px;
+    font-weight: 600;
+    color: #fff;
+    cursor: pointer;
+    margin: 20px;
+    height: 55px;
+    text-align:center;
+    border: none;
```
asynchroza, line 27: Do you need border `none` ? If not, remove it. Our goal is to ship the least amount of "bytes" and applicable styles from the server to the browser.

asynchroza (changes requested):

asynchroza:
@DemonHashira any updates?

asynchroza:
Button is breaking / looking weird on IOS Safari. This should be investigated.

Furthermore, I think the text is too bold and the button is too rounded. It's not matching the style of the desktop buttons.

Issue #113, opened 2022-12-10
Fix spacing between Navbar buttons in Desktop


SuelaCark:
Add more space between anchors when width: 900px
Remove space between anchors while on full desktop width.

asynchroza:
I've disabled the HackAUBG button until further notice. We should focus on creating the HackAUBG page.

PR #117, opened 2022-12-10
#113 less space between buttons

SuelaCark:
#113 Reduced space between anchor tags and button


asynchroza:
@IvanObreshkov , @SuelaCark has already made some changes but please, when you work on the issue where you will be adding the anchors from the hubconf page, put the anchors even closer to each other.

Issue #97, opened 2022-11-30
Add min-height property to admin panel


asynchroza:
## Brief description:
<img width="427" alt="Screenshot 2022-11-30 at 10 25 12" src="https://user-images.githubusercontent.com/104720011/204744942-dd76a016-cc0a-4af1-9c18-22f88b173fea.png">

Should be:
<img width="462" alt="Screenshot 2022-11-30 at 10 29 22" src="https://user-images.githubusercontent.com/104720011/204745850-84c208dd-5089-4a82-aca1-3366b27d3c6c.png">

## How to achieve it:
Apply a min-height: 100vh property to the correct element

asynchroza:
@annakukova please, help him out when you have the time

asynchroza:
Issue is persistent on desktop versions as well:
![image](https://user-images.githubusercontent.com/104720011/208712605-0c6a7c1d-862c-4160-9f2e-f92d383afd10.png)

PR #154, opened 2022-12-22
#97 add min-height property to admin panel 

Aleks1908:
Added min-height property to the admin page. issue:#97

### DIFF
--- a/packages/web/src/components/admin_page/admin.css
+++ b/packages/web/src/components/admin_page/admin.css
@@ -53,7 +53,7 @@
 
 /* dashboard */
 .dash {
-    height: 100vh;
+    min-height: 100vh;
     display: flex;
     flex-direction: column;
     justify-content: center;
@@ -171,7 +171,7 @@
 
 .add-member-main-div {
     color: #fff;
-    height: 100vh;
+    min-height: 100vh;
     padding-top: 2rem;
     display: flex;
     flex-direction: column;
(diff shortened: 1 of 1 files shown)


asynchroza:
image
--- 
there is a bug with the flex, members and events become unreachable at certain resolutions
image

asynchroza:
@Aleks1908 furthermore, the issue reference should happen in the beginning of the title. Please, refer to old PRs and the Wiki.

packages/web/src/components/admin_page/admin.css
```diff
@@ -205,9 +205,6 @@
 }
 /* respnsive */
 @media only screen and (max-width: 414px) {
```
asynchroza, line 207: why is it 414px specifically?

packages/web/src/components/admin_page/admin.css
```diff
@@ -104,7 +104,7 @@
     display: flex;
     flex-direction: column;
     background-image: linear-gradient(rgb(21, 76, 121), rgb(57, 95, 105));
-    height: 100vh;
+    min-height: 100vh;
```
asynchroza, line 107: should be `min-height: 100vh` + `height: 100%` so that it expands

asynchroza (changes requested):

asynchroza:
@Aleks1908 good job!

PR #157, opened 2022-12-23
#156 Added make to the script

dankneebee:


asynchroza (approved):
Great job Dan! I have forgotten about that!

Issue #107, opened 2022-12-09
Landing animation


asynchroza:
## Brief description:
Create some cool landing animation (e.g. a text which gets typed and it says "Welcome to The Hub" and then the screen fades out)
## How to achieve it:
IDK, u know

asynchroza:
@SuelaCark can you make the cursor stop immediately after the `Welcome to The Hub` text.

PR #158, opened 2022-12-23
#107 landing animation

SuelaCark:
Made the Landing animation for the "Welcome to the Hub" section and Hubzzie.
Fixed responsiveness for mobile.


asynchroza:
Has the animation always been like that? It is stuttering too much and I feel like it wasn't like that before.

Issue #172, opened 2023-01-28
About us section


SuelaCark:
Create the About us section right after the Welcome to the Hub section. Please also work on the following:

- The responsiveness of the about us section (use flexbox and media queries. For this you can checkout the css of the "welcome to the hub" section as an example)
- Connect the "About us" anchor tag with the "About us" section title. For this, you can checkout the hubconf webpage anchors sliding effect.

The template design can be found here: [SPA website design](https://xd.adobe.com/view/0a9f9bd7-20d6-47c0-b63d-3766ef3c8767-5fcc/screen/e50a5fa8-0255-4c9a-9d36-41ca3d819bae/)

PR #185, opened 2023-02-04
#172 about us section

Aleks1908:
#172 Created about us section


SuelaCark (approved):

asynchroza (changes requested):
Remove all package*.json files which are outside of the scope of the `packages/web` directory. Remove `node_modules` as well!

PR #188, opened 2023-02-04
#169 Fixed nav expanding

Aleks1908:
Fix bug where nav wasn't expending to the page's height when an anchor is clicked


asynchroza (commented):
LGTM

Issue #195, opened 2023-02-04
Change Hubzzie image


SuelaCark:
Change Hubzzie image in the main landing page

PR #199, opened 2023-02-04
#195 Changed hubzie image

Eldi47:
#195 Changed the hubzzie image and changed the "About us" text to "Meet the team"


packages/web/src/components/spa/LandingSection/landing_section.css
```diff
@@ -66,13 +66,23 @@
 
 .hubzzie {
     width: auto;
-    align-items: center;
-    margin-right: 15%;
 }
 
-@media only screen and (min-width: 100px) {
+@media only screen and (min-width: 200px) {
```
asynchroza, line 71: Why is this `200px`?

Eldi47: Tried to work with min-width values for the image to appear the correct size. Will fix this issue in a sec.

asynchroza: If you think of leaving this in it would be a good idea to change `200px` -> `400px` or some bigger resolutions where you can expect breaking changes in the positioning of elements (e.g. Samsung Fold resolution)

Eldi47: So you are suggesting keeping the `100px` one and create a `400px` for  bigger devices?

asynchroza: No, I suggest you swap the `100px` one for the `400px`

asynchroza: Never mind that, code works as expected so no further improvements are needed!

asynchroza (approved):

PR #237, opened 2023-02-07
#175 hackaubg stepsjourney section

Aleks1908:
Create Journey Section for mobile and desktop


packages/web/src/components/spa/HackAUBG/JourneySection/JourneyMobile/JourneyMobileOld.jsx
```diff
@@ -0,0 +1,65 @@
+import React from 'react';
```
asynchroza, line 1: Remove old file

packages/web/src/components/spa/HackAUBG/JourneySection/JourneyMobile/JourneyStep.jsx
```diff
@@ -0,0 +1,58 @@
...
+    const [stepContent, setStepContent] = useState(
+        'container-content-not-displayed'
+    );
+    const [arrowDown, setArrowDown] = useState('journey-arrow-displayed');
+    const [arrowUp, setArrowUp] = useState('journey-arrow-not-displayed');
+    const getHeight = () => {
+        const divElement = document.getElementById('test');
+        if (divElement) {
+            return divElement.scrollHeight;
+        } else {
```
asynchroza, line 17: you don't need the else statement and return here

packages/web/src/components/spa/HackAUBG/JourneySection/JourneyMobile/journey_mobile_old.css
```diff
@@ -0,0 +1,34 @@
...
+    flex-direction: column;
+    align-items: center;
+    justify-content: center;
+}
+.journey-mobile-content {
+    display: flex;
+    flex-direction: column;
+    align-items: center;
+}
+.journey-mobile-title {
```
asynchroza, line 12: Add newline between classes

packages/web/src/components/spa/HackAUBG/JourneySection/JourneySection.jsx
```diff
@@ -0,0 +1,13 @@
+import { JourneyDesktop } from './JourneyDesktop/JourneyDesktop';
+import { JourneyMobile } from './JourneyMobile/JourneyMobile';
+import { useMediaQuery } from 'react-responsive';
+
+export const JourneySection = ({ props }) => {
+    const isMobile = useMediaQuery({ query: '(max-width: 1200px)' });
```
asynchroza, line 6: test whether it works with 1000px. This is the breakpoint we usually use. If it doesn't, leave it like that

packages/web/src/components/spa/HackAUBG/HackAUBG.jsx
```diff
@@ -30,12 +29,14 @@ export const HackAUBG = () => {
                         'rgba(0,0,0,.5)',
                         true,
                         '#222222',
-                        'red'
+                        'red',
+                        1
```
asynchroza, line 33: You don't need the 1 here

asynchroza (changes requested):

packages/web/src/components/spa/HackAUBG/JourneySection/JourneySection.jsx
Aleks1908: I know we use 1000 but the design with the lines and dots does not look ok on 1000 and i think it will be better if it renders the mobile one on 1200

packages/web/src/components/spa/HackAUBG/JourneySection/JourneyMobile/journey_mobile_old.css
Aleks1908: this is about the old css that has been removed

packages/web/src/components/spa/HackAUBG/JourneySection/JourneySection.jsx
asynchroza: Ok, that's fine.

asynchroza:
Can we move the dots to be on the same line as the headers? I think the current position of the dots is really weird.

asynchroza:
Otherwise, the mobile version is amazing! Great job on that!

SuelaCark:
> Can we move the dots to be on the same line as the headers? I think the current position of the dots is really weird. DM me when you are ready with this, send me a pic of the refreshed component and I will approve.

^I agree with @asynchroza about the dots on the same level with the headers. In desktop version, I would also make the last line a little longer so it can look like its connecting with the other section and not like its being cut. Otherwise the mobile version is perfect and has a smooth transition. Well done! @Aleks1908

SuelaCark (dismissed):
LGTM!

SuelaCark:
On mobile its fine, but when you are on responsiveness (desktop < 900px), you can see that the text of the Step box shows before the animation can fully open.

asynchroza (approved):

Issue #250, opened 2023-02-08
Hamburger Menu bug while closing


SuelaCark:
Nav inner elements should be hidden when a scrolling (after closing). Everything inside the nav disappears

PR #255, opened 2023-02-10
#250 Fix hamburger menu elements disappearing when closed

Eldi47:
When the menu is set to navmobile-menu-backwards, the menu becomes scrollable, stretches to the entire page, and the anchors go to the center of the extended page, outside the visible display.  To fix this, I removed the flex display and added a top margin so that the anchors stay fixed when the menu stretches to the end.


asynchroza (changes requested):
Great job, Eldi! This is a neat fix! 

Can you please add a z-index >= 7 to the mobile nav because whenever you click on team, the buttons of the team carousel can be seen before the nav has fully hidden itself.

packages/web/src/components/spa/Navigation/MobileNav/mobile_navbar.css
```diff
@@ -1,6 +1,7 @@
 @import url('https://fonts.googleapis.com/css2?family=Poppins:wght@200;300;400;500;600;700&display=swap');
```
asynchroza, line 1: Can you remove this line?

asynchroza (approved):

Issue #254, opened 2023-02-10
Make Video section responsive for Iphone X and Samsung Fold


SuelaCark:
The responsiveness of the Video section is okay for most mobile devices, however it can be modified. Particularly on Iphone X and Samsung Fold, the video section breaks the page.
Hence, it would be easier to work on fixing the responsiveness for all mobile devices and not only for particular ones.

PR #256, opened 2023-02-11
#254 Video Section responsive to all resolutions

Eldi47:
The video width and height now adapt to all resolutions. Added bottom padding to the video section and increased the section title font size for smaller resolutions.


packages/web/src/components/spa/HackAUBG/VideoSection/video_section.css
```diff
@@ -15,6 +15,8 @@
     align-items: center;
     display: flex;
     flex-direction: column;
+
+    padding-bottom: 20px;
```
SuelaCark, line 19: Please change padding to margin and use percentages (%) instead of pixels (px)

SuelaCark (changes requested):
The text and video responsiveness LGTM. Just change the padding to margin and please check to delete and not include the following:
image

SuelaCark:
@Eldi47 When you are working on the things in the above-mentioned comments, please also work on the following:
(1) Resizing the video for desktop into having more width. We want the video width to be similar to the one from the webpage https://thehub-aubg.com/hackaubg (also check pictures below for reference)
(2) Add height to the overall container. At times, the unassigned height is leading to other sections getting into the video section space.

new video section
old video section

packages/web/src/components/spa/HackAUBG/VideoSection/video_section.css
```diff
@@ -24,20 +27,28 @@
...
     align-items: center;
     display: flex;
-    padding: 3% 0;
     padding: 3% 0%;
     text-align: justify;
 }
 
 @media only screen and (max-width: 900px) {
+    .video-container {
+        height: 90vh;
```
SuelaCark, line 36: This height for mobile is a little bit too much. There is too much space at the top and bottom part between the video section and the other sections.

packages/web/src/components/spa/HackAUBG/VideoSection/video_section.css
```diff
@@ -24,20 +27,28 @@
...
+        text-align: center;
     }
     .video {
         width: 80%;
         justify-content: center;
         align-items: center;
         display: flex;
     }
+
+    .video iframe {
```
SuelaCark, line 50: Isn't this suppose to be equivalent to .video div? Since the div video is only used for holding the iframe, you can remove the div and add the classname to the iframe tag, then move all css from iframe to the .video class.

packages/web/src/components/spa/HackAUBG/VideoSection/video_section.css
```diff
@@ -24,20 +27,28 @@
...
 
 @media only screen and (max-width: 900px) {
+    .video-container {
+        height: 90vh;
+    }
+
     .video-header {
-        text-size: 0.5rem;
-        text-align: justify;
+        font-size: 1.5em;
```
SuelaCark, line 40: Check the font-size immediate change between 1200px and 1300px. Please make the font-size bigger for devices smaller than 1300px

SuelaCark (changes requested):

SuelaCark:
@Eldi47 please work on the requested changes. Also note that I removed the .video div and moved the class to the iframe, since the div was just holding the iframe and I didnt see its use as necessary.

packages/web/src/components/spa/HackAUBG/VideoSection/video_section.css
```diff
@@ -23,20 +20,38 @@
...
+        width: 100%;
+    }
+}
+
+@media only screen and (max-width: 1300px) {
+    .video-header {
+        font-size: 2.5em;
+    }
+}
+
```
Aleks1908, line 52: Why are there two media queries with the same max-widths. You can combine them in one. And also for some reason the text is too big and not responsive on mobile devices.

Aleks1908 (changes requested):

packages/web/src/components/spa/HackAUBG/VideoSection/video_section.css
```diff
@@ -23,20 +20,38 @@
...
+@media only screen and (min-width: 1300px) {
```
Aleks1908, line 53: Why are there two media queries with the same max-widths. You can combine them in one. And also for some reason the text is too big and not responsive on mobile devices.

Aleks1908 (changes requested):

packages/web/src/components/spa/HackAUBG/VideoSection/video_section.css
SuelaCark: @Aleks1908 If you look closely you can see that one media query is used for max-width until 1300px (devices with width lower than 1300px), and the other is used for devices with width bigger than 1300px. However, I do agree that @Eldi47 should work on creating better media queries for implementing responsiveness. FYI, the default media query that we are using for mobile devices is max-width 900px.

SuelaCark (approved):
LGTM.

SuelaCark (approved):

Issue #201, opened 2023-02-04
HackAUBG Desktop Navbar spacing


SuelaCark:
The desktop navbar of HackAUBG on 900px-1000px the anchors don't have enough spacing between them.

SuelaCark:
@Pessy1 Please look at the following picture as a guide on what should be fixed. The Grading criteria while on 900-1050px has too much space on the right. Please use media queries to fix this issue. If you need further instruction just let me know. 

![image](https://user-images.githubusercontent.com/77395265/217548146-8e772ea4-db6f-4e95-ae3f-26f6a6cfa9bf.png)

PR #258, opened 2023-02-11
#201 added media query to fix navbar spacing

Pessy1:
Media query added for the  "Grading criteria" spacing issue on the desktop navbar.


SuelaCark (changes requested):
This change is breaking both Navbars (landing page + HackAUBG - since they are connected) while on responsiveness. The nav anchors should be held within their div while on responsiveness and not exceed it. Please check the following pictures as a reference for what its breaking.

image
image

SuelaCark (approved):
LGTM. Good job @Pessy1

Issue #266, opened 2023-02-13
404 page logo size


SuelaCark:
On the 404 page, make the Hub logo bigger and 'the HUB' text a little smaller. Please ensure that the proportion between the two elements is normal and acceptable. 

![image](https://user-images.githubusercontent.com/77395265/218488929-bfd2ed50-b787-4fae-8188-616aba72a2bd.png)

PR #285, opened 2023-02-17
#266 Changed image and text values on 404 Page 

DemonHashira:


asynchroza:
I feel like the page is not spaced out evenly:

image


image

@SuelaCark @Aleks1908 @annakukova weigh in. idk if this is important enough to be addressed rn.

Aleks1908:
About the spacing @DemonHashira you can try playing with margin-top for desktop and mobile. While you need to make sure that they are in % and to have a media query for mobile if it does not work with only one.

SuelaCark:
> About the spacing @DemonHashira you can try playing with margin-top for desktop and mobile. While you need to make sure that they are in % and to have a media query for mobile if it does not work with only one.

@DemonHashira Please review the requested changes by @asynchroza and also review the suggested solution by Aleks. Additionally, while you are working on those, I think you can also center the <p> text of the 404page. As you can see in the image below, the paragraph text is not centered within itself, but it has left-alignment (by default). 

image

SuelaCark (changes requested):
@DemonHashira Please look at the conversations and attached screenshots for reference and make the relevant changes.

nikolayninov (approved):

asynchroza:
@SuelaCark @Aleks1908 can you validate this? It seems that the other PR is conflicting with master because this one is not merged in master and is based on this one.

SuelaCark (changes requested):
Generally LGTM. However, one thing that I want you to fix is to add more space between the logo and the hubzzie-UFO image. There is not enough space there (this happens in particular devices: Iphone SE and Samsung Galaxy S8+).

image

packages/web/src/components/other/page_not_found.css
```diff
@@ -94,8 +94,8 @@
     .hubzie-UFO-image {
         width: 100%;
         justify-content: center;
-        margin-top: 5%;
-        margin-bottom: 5%;
+        margin-top: 10%;
+        margin-bottom: 1%;
```
SuelaCark, line 99: Everything LGTM, just please add a little more margin bottom for the hubzie-UFO-image because in all devices, the 404 text seems too close to the image. After being done please resolve conversations and re-request review from me.

SuelaCark (changes requested):

packages/web/src/components/other/page_not_found.css
```diff
@@ -85,13 +89,14 @@
...
     }
     .hubzie-UFO-image {
         width: 100%;
         justify-content: center;
-        margin-bottom: 3%;
+        margin-top: 11%;
+        margin-bottom: 4%;
     }
     .page-not-found-text {
         width: 100%;
```
SuelaCark, line 102: Here add margin-right: 0; On mobile the whole text div is not properly centered because it inherits margin-right: -5% from line 48. Everything else is fine. Again, after making the requested changes don't forget to resolve conversations and re-request review. 

image

SuelaCark (changes requested):

SuelaCark:
Everything LGTM. However, I dont know if we should check the resizing of the page for Iphones while using Safari since Safari has the url link and search query positioned in the bottom (this causes the Back to Homepage button to be cropped).
Also, @DemonHashira we should probably include a onHover effect for the Back to Homepage button (just get the same animation as Meet the team button in the landing page).

Issue #295, opened 2023-02-19
Footer Icons issue on mobile devices


Aleks1908:
React icons look stretched on mobile devices

PR #296, opened 2023-02-19
#295 footer icons issue on mobile devices

Aleks1908:
Added spans to all icons and remade borders

### DIFF
--- a/packages/web/src/components/spa/Footer/Footer.jsx
+++ b/packages/web/src/components/spa/Footer/Footer.jsx
@@ -12,30 +12,37 @@ export const Footer = ({ colour, iconcolor }) => {
             </div>
             <div className="social-media">
                 <div className="links">
-                    <a
-                        href="https://www.facebook.com/TheHubAUBG"
-                        target="_blank"
-                    >
-                        <FaFacebookSquare
+                    <span className="icon-footer">
+                        <a
+                            href="https://www.facebook.com/TheHubAUBG"
+                            target="_blank"
                             style={{ backgroundColor: iconcolor }}
-                        />
-                    </a>
+                        >
+                            <FaFacebookSquare />
+                        </a>
+                    </span>
                 </div>
                 <div className="links">
-                    <a
-                        href="https://www.instagram.com/thehubaubg/"
-                        target="_blank"
-                    >
-                        <GrInstagram style={{ backgroundColor: iconcolor }} />
-                    </a>
+                    <span className="icon-footer">
+                        <a
+                            href="https://www.facebook.com/TheHubAUBG"
+                            target="_blank"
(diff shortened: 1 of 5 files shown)


packages/web/src/components/spa/Footer/Footer.jsx
```diff
@@ -12,30 +12,37 @@ export const Footer = ({ colour, iconcolor }) => {
             </div>
             <div className="social-media">
                 <div className="links">
-                    <a
-                        href="https://www.facebook.com/TheHubAUBG"
-                        target="_blank"
-                    >
-                        <FaFacebookSquare
+                    <span className="icon-footer">
```
asynchroza, line 15: Instead of an anchor element,bind the function for opening links from Global.js to an onClick on the span component.

packages/web/src/components/spa/Footer/Footer.jsx
```diff
@@ -12,30 +12,37 @@ export const Footer = ({ colour, iconcolor }) => {
...
+                        <a
+                            href="https://www.facebook.com/TheHubAUBG"
+                            target="_blank"
                             style={{ backgroundColor: iconcolor }}
```
asynchroza, line 19: I don't remember what was the idea of this style property. Is this for the background or the icons themselves? Probably we should have two properties, one for the background and one for the icon's actual Color. Furthermore, can you rename the iconcolor prop to iconColor? We should stick to camelCase.

packages/web/src/components/spa/Footer/Footer.jsx
```diff
@@ -3,39 +3,47 @@ import './footer.css';
...
                 </div>
                 <div className="links">
-                    <a
-                        href="https://www.youtube.com/channel/UChdtBZBvaK9XZurP3GjPDug"
-                        target="_blank"
+                    <div
+                        className="footer-icon"
+                        style={{ backgroundColor: iconBgColor }}
+                        onClick={() => {
+                            openNewTab('https://www.facebook.com/TheHubAUBG');
```
asynchroza, line 42: This should be YouTube

packages/web/src/components/spa/Footer/Footer.jsx
```diff
@@ -3,39 +3,47 @@ import './footer.css';
...
                 </div>
                 <div className="links">
-                    <a
-                        href="https://www.instagram.com/thehubaubg/"
-                        target="_blank"
+                    <div
+                        className="footer-icon"
+                        style={{ backgroundColor: iconBgColor }}
+                        onClick={() => {
+                            openNewTab('https://www.instagram.com/thehubaubg/');
```
asynchroza, line 31: Lose the www subdomain

SuelaCark (approved):
LGTM

PR #312, opened 2023-02-25
#305 Landing sections to be full screen

Aleks1908:
Changed view heights of the landing pages so that they are on full screen


asynchroza:
Can you add this animation? https://www.youtube.com/watch?v=W5oawMJaXbU , we've discussed this as the part of the ticket this PR is for but I guess we had forgotten to add it to the issue.

Aleks1908:
> Can you add this animation? https://www.youtube.com/watch?v=W5oawMJaXbU , we've discussed this as the part of the ticket this PR is for but I guess we had forgotten to add it to the issue.

I should add this animation only on the HackAUBG page right?

asynchroza:
> > Can you add this animation? https://www.youtube.com/watch?v=W5oawMJaXbU , we've discussed this as the part of the ticket this PR is for but I guess we had forgotten to add it to the issue.
> 
> I should add this animation only on the HackAUBG page right?

Yes indeed

packages/web/src/components/spa/HackAUBG/LandingAnimation/landing_animation.css
```diff
@@ -56,7 +60,7 @@
     }
     .animation-container {
         padding-top: 6vh;
-        height: 45vh;
+        height: 100vh;
```
SuelaCark, line 63: If you make the height: 100vh in the media query max-width: 1300px, you dont have to do the same for the other media queries since this attribute already applies to all of them (because they are lower than 1300px)

packages/web/src/components/spa/HackAUBG/LandingAnimation/landing_animation.css
```diff
@@ -72,14 +76,14 @@
     }
     .animation-container {
         padding-top: 7vh;
-        height: 45vh;
+        height: 100vh;
```
SuelaCark, line 79: Same here and below. Please remove the height:100vh from all media queries except for the max-width:1300px (because this css will make no difference)

SuelaCark (changes requested):
Look at the comments about the media queries and make appropriate changes. 

**General note:** please come up with simpler names for your branches. I got a heart attack just by trying to read it.

packages/web/src/components/spa/HackAUBG/LandingAnimation/landing_animation.css
```diff
@@ -72,14 +76,14 @@
     }
```
SuelaCark, line 76: Also, make the font-size for mobile devices bigger since now the font is too small and we can cover more space due to the increase of the animation space.

asynchroza:
> Look at the comments about the media queries and make appropriate changes.
> 
> **General note:** please come up with simpler names for your branches. I got a heart attack just by trying to read it.

this is automatically generated from github

packages/web/src/components/spa/HackAUBG/LandingAnimation/landing_animation.css
asynchroza: we don't need any media queries for setting height at all if the height is always 100vh

packages/web/src/components/spa/HackAUBG/LandingAnimation/landing_animation.css
asynchroza: Remove height from max-width as well

packages/web/src/components/spa/HackAUBG/LandingAnimation/LandingAnimation.jsx
```diff
@@ -1,11 +1,48 @@
...
+
+    useEffect(animation, []);
+
     return (
         <div className="animation-container">
             <div className="landing-content">
-                <h1>HackAUBG 5.0</h1>
+                <h1
+                    data-value="HACKAUBG 5.0"
+                    className="toni-montana"
```
asynchroza, line 41: change the className to something more descriptive

packages/web/src/components/spa/HackAUBG/LandingAnimation/LandingAnimation.jsx
```diff
@@ -1,11 +1,48 @@
...
+            }
+
+            iteration += 1 / 3;
+        }, 30);
+    };
+
+    useEffect(animation, []);
+
     return (
         <div className="animation-container">
```
asynchroza, line 37: if you are not using moduled css, you should be more descriptive with your css classes. For example, here the className should probably be "hackaubg-landing-section-text-animation-container"

packages/web/src/components/spa/HackAUBG/LandingAnimation/LandingAnimation.jsx
```diff
@@ -1,11 +1,48 @@
...
             <div className="landing-content">
```
asynchroza, line 38: here as well

packages/web/src/components/spa/HackAUBG/LandingAnimation/LandingAnimation.jsx
```diff
@@ -1,11 +1,48 @@
 import React from 'react';
+import { useEffect } from 'react';
 import './landing_animation.css';
 
 export const MatrixWindow = () => {
+    const animation = () => {
```
asynchroza, line 6: probably rename this to `animateLetters`

packages/web/src/components/spa/HackAUBG/LandingAnimation/LandingAnimation.jsx
asynchroza: @Aleks1908 className still seems to be `toni-montana`

Issue #314, opened 2023-02-26
Update schedule section text


annakukova:
**The updated schedule is the following:**

Note: This schedule is prone to changes. Hence, check with @SuelaCark before approving and merging PRs that include this ticket.

_**NEW CHANGES IN SATURDAY SCHEDULE**_

Friday:

17:30 Registration
18:30 Opening 
20:00 Brainstorming
21:00 Idea Pitching
21:30 Start Coding

Saturday:

9:30 Breakfast
11:00 Mentor & HR Meetings
11:30 Idea Pitching Pt. 2
12:00 Mentor & HR Meetings
13:30 Lunch
14:30 Mentorship + Coding
19:30 Dinner
20:30 Coding

Sunday:

9:30 Breakfast
10:30 Coding
13:00 Presentations Submission Deadline
13:00 Lunch
(right after they submit the presentations they have lunch yes)
14:30 Presentations Begin
19:00 Award Ceremony

savinadd:
I have changed the layout to either a horizontal row with three columns or one column with three rows depending on screen width, @annakukova and I discussed it and we think it would look better this way. @SuelaCark @annakukova, are we sticking to the idea of green lines between the days that @SuelaCark suggested? in my opinion, i think it might make the grading criteria and the schedule section look very similar in design

annakukova:
It looks nice and clean to me as it is right now, but if we do decide to have the lines in between, I would suggest that they're only for the desktop version. Responsive looks nice and the headings for each day act as separators so IMO no need of them for mobile.

PR #316, opened 2023-02-26
#314 updated schedule text

annakukova:
#314 Updated the text in the schedule section


SuelaCark:
@annakukova or @savinadd Please work on the following changes for the schedule section (review pictures below as reference):
(1) Make the font-size of the text bigger, since now it looks insignificant.
(2) Have less space between columns since they are too separate from each-other.
(3) Responsiveness on mobile --> to what I have noticed there is no particular media query that deals with mobile responsiveness (except for one media query targeting devices with width < 350px)


image
image

SuelaCark (changes requested):
@savinadd Can you please work again on this ticket? I don't really like how the schedule section looks and it lacks good responsiveness. Please try to work more on the text alignment and its responsiveness because to what I can see, the texts in different days are not aligned within the same row. It looks like Saturday and Sunday are aligned lower than Friday (kinda looking like stairs --> refer to pictures below). Also, there are new changes to the schedule from the board so please refer to the #314 ticket to get the new changes. 

I was thinking of adding a design where we have the days divided by a green line. What do you think of implementing such design? @annakukova @savinadd 

image
image

SuelaCark (changes requested):
(1) The space between saturday and sunday is much less then the space between friday and saturday. You can notice this at width a little above 1100px.
image

(2) Below width 900px there is a lot of space in one row for 2 columns, yet there are only 1 column per row. Also, the columns are itself justified in the left. Here, center the columns themselves in the schedule container, don't center their text. 
image

(3) Add more padding-top between the Schedule header and the video section in mobile

(4) As per the addition of green lines design, I believe that it would be a good compliment to the HackAUBG page since pretty much the other sections (journey and grading criteria) include such similar theme/aesthetic. Hence, please try to add these designs in this section (don't forget about how they will look in mobile devices and during responsiveness).

SuelaCark (approved):

Issue #317, opened 2023-02-27
Make HackAUBG letters animation slower and make journey section cards more rounded


Aleks1908:

asynchroza:
Can you as well change the font and/or the boldness of the text below 
<img width="725" alt="image" src="https://user-images.githubusercontent.com/104720011/221562503-08754c62-21ed-4754-b7a8-f7ce6c8de29d.png">

Also it seems that the font is a bit small when viewed from a phone.

PR #318, opened 2023-02-27
#317 make hackaubg letters animation slower

Aleks1908:
Changed font-sizes and removed unnecessary code and changed border radius for journey section mobile


packages/web/src/components/spa/HackAUBG/LandingAnimation/landing_animation.css
```diff
@@ -33,28 +33,17 @@
...
-        font-size: 3.8rem;
-    }
-    .hackaubg-landing-content p {
-        font-size: 0.9rem;
-        padding: 2% 0 2% 0;
-        text-overflow: clip;
-    }
+@media only screen and (max-width: 900px) {
     .hackaubg-landing-section-text-animation-container {
         height: 100vh;
```
asynchroza, line 45: isn't it supposed to always be 100vh? why is this set in the media-query and not at a top level?

packages/web/src/components/spa/HackAUBG/LandingAnimation/landing_animation.css
```diff
@@ -33,28 +33,17 @@
 }
 
 .hackaubg-landing-content p {
-    font-size: 1rem;
+    font-size: 1.2rem;
```
asynchroza, line 36: I can't run it locally rn but my guess is that it's still a bit small. @SuelaCark probably should check up on that. In my opinion you should increase the font-weight as well.

SuelaCark (changes requested):
I think it would be better if you could make the font-size bigger (also for the <p> text) for mobile because the Go to Registration button font and size is bigger than the text above. 

Side note: I noticed that at times the page (especially font-type and the gif animation) is taking too much time to load. While, the text animation takes place immediately

asynchroza (approved):

Issue #327, opened 2023-03-03
Mobile text margins


SuelaCark:
Make all texts in mobile have the same left and right margins.

PR #329, opened 2023-03-03
#327-HackAUBG-Mobile-Text-Margins

r-mishev:
#327 Made margins consistent in responsive


packages/web/src/components/spa/HackAUBG/GradingCriteria/grading_criteria.css
```diff
@@ -48,7 +48,7 @@
 @media only screen and (max-width: 900px) {
```
asynchroza, line 48: On desktop there is more padding/margin on the top than on the bottom. Make it the same for both desktop and mobile devices

Issue #326, opened 2023-03-01
Journey section mobile font-size


SuelaCark:
Make font-size bigger for the journey section on mobile devices with width less than 1200px. 

![image](https://user-images.githubusercontent.com/77395265/222287423-a0c6961d-7a28-4725-b860-e58b0aacaaee.png)

PR #330, opened 2023-03-03
#326 Made journey section font size bigger

Eldi47:
The font size for the journey section appeared too small for devices 1200px and lower so I changed the `font-size:` to be in `vw` so that the text size is consistent in all width confirgurations.


asynchroza (changes requested):
Text is too big now, it should be somewhere in between the current change and the previous state

SuelaCark:
@Eldi47 update your branch with master (do this in your LOCAL branch and push the changes, not the shortcut of github). If you need guidance, read the [wiki (click here).](https://github.com/AUBGTheHUB/spa-website-2022/wiki/Pull-Requests-and-Updating-branches-with-master)

SuelaCark (changes requested):
Text for mobile devices with width from 600-1200px is too big. Try using font-size with rem/em instead of vw

SuelaCark (changes requested):
Now the font is too big for mobile. Try to work with something similar to what you had worked 1 commit before, but use media queries to make font-adjustments.

image

SuelaCark:
@Aleks1908 @asynchroza @nikolayninov Please review the changes made and approve if you think that the font-sizes are better adjusted now. If there needs to be any additional changes, please ping @Eldi47

Aleks1908 (approved):

Issue #328, opened 2023-03-03
Add Scrollbar design


SuelaCark:
As of now the website's scrollbar does have a design. The initial idea is to make the slider have rounded edges, black background and the scrollbar to have a green color. Checkout the design of the last scrollbar from [this link](https://i.pinimg.com/originals/82/3d/e2/823de2fb01f5c64f7778d0a1d69121e4.jpg)  as a design reference.

PR #331, opened 2023-03-03
#328 Added new scrollbar design

Eldi47:
Made the scrollbar thumb color green with the background color black.


asynchroza (changes requested):
Scrollbar should only be available on HackAUBG page

SuelaCark (approved):
LGTM!

Issue #232, opened 2023-02-06
Add Hub Logo Favicon to the website


nikolayninov:

Mohammad-Farhan64:
<!DOCTYPE html>
<html lang="en">
<head>
    <meta charset="UTF-8">
    <meta http-equiv="X-UA-Compatible" content="IE=edge">
    <meta name="viewport" content="width=device-width, initial-scale=1.0">
    <title>Add Hub Logo Favicon to the website · Issue #232 · AUBGTheHUB/spa-website-2022</title>

    <link rel="icon" class="js-site-favicon" type="image/svg+xml" href="https://github.githubassets.com/favicons/favicon-dark.svg"></head>
<body>
   
</body>
</html>

Mohammad-Farhan64:
<title>Add Hub Logo Favicon to the website · Issue #232 · AUBGTheHUB/spa-website-2022</title>

      <link rel="icon" class="js-site-favicon" type="image/svg+xml" href="https://github.githubassets.com/favicons/favicon-dark.svg">

PR #336, opened 2023-03-03
#232 Add new logic for favicon change

asynchroza:
@annakukova can you make the green favicon be the same size as the blue one and replace all of the android/iphone favicons with new ones


SuelaCark (approved):
LGTM

Issue #291, opened 2023-02-19
Remove margin-bottom from Awards section main div


SuelaCark:
The main div awards-section-hack-aubg-container has a margin-bottom which gives too much space between the awards section and the FAQ section. Please remove this margin, also change the name of the div container since its too long and complicated (this can be avoided). 

Don't forget that if you change the name of the div in the jsx file, you must also change its name in the css!

![image](https://user-images.githubusercontent.com/77395265/219945538-54a6aa0c-b4ba-4a1e-bddb-9b2cd1257a46.png)
![image](https://user-images.githubusercontent.com/77395265/219945556-345b6f91-d1bc-441c-9c70-d00fc7108328.png)

PR #345, opened 2023-03-04
#291 Simplified class names for Award Section

Eldi47:
Simplified the div container names in the Award Section.


SuelaCark (changes requested):
The images now are too big on desktop! @Eldi47 please fix this issue

image

SuelaCark (approved):

Issue #323, opened 2023-03-01
HackAUBG background colors


SuelaCark:
The white background sections need to change to gray and add the gradient effect (same as the video section) to some of the sections.

PR #346, opened 2023-03-04
#323 hack aubg background colors

Pessy1:
Used a dark green colour for beckground and lighter green for text


SuelaCark:
Note here that I remove commented css codes from all files and at the same time fixed another issue which was the breakage of the awards section text for mobile devices.

SuelaCark (approved):

Issue #349, opened 2023-03-04
Integrate mailing service into hackathon_register_controller


IvanObreshkov:
## Brief description:
Integrate mailing service into hackathon_register_controller.go 
## How to achieve it:
Create a function to make a request to the mailing service when a new participant has been created and added to the DB

PR #360, opened 2023-03-05
#349 Integrate mailing service in registration

asynchroza:


packages/api/controllers/hackathon_register_controller.go
```diff
@@ -152,3 +161,49 @@ func FormatTeamName(teamName string) string {
...
+
+	type Mailer struct {
+		Receiver string `json:"receiver"`
+		Html     string `json:"html"`
+		Subject  string `json:"subject"`
+	}
+
+	var reqBody Mailer
+
+	reqBody.Html = "<h1>HELLO</h1>"
```
asynchroza, line 177: This should be a dynamically loaded template. I'm attaching last year's template here: https://github.com/AUBGTheHUB/website-2021/blob/3e5acfc89cd27b529da295748805838af42b1314/packages/api/src/templates/confirmation.html

Instead of "Dear Participant", make it "Dear {name}"

packages/api/controllers/hackathon_register_controller.go
```diff
@@ -152,3 +161,49 @@ func FormatTeamName(teamName string) string {
...
+	reqBody.Receiver = email
+	reqBody.Subject = "TEST"
```
asynchroza, line 179: Subject should probably be `Welcome to HackAUBG 5.0` or something like that.

packages/api/controllers/hackathon_register_controller.go
```diff
@@ -152,3 +161,49 @@ func FormatTeamName(teamName string) string {
...
+
+	fmt.Println(reqBody)
+	json_data, _ := json.Marshal(reqBody)
+
+	client := &http.Client{}
+
+	req, err := http.NewRequest(http.MethodPost, requestURL, bytes.NewBuffer(json_data))
+
+	if err != nil {
+		fmt.Println("could not create request to mailing service: ")
```
asynchroza, line 189: Why do you keep going if the request creation failed? exit function

packages/api/controllers/hackathon_register_controller.go
```diff
@@ -152,3 +161,49 @@ func FormatTeamName(teamName string) string {
...
+		fmt.Print(err.Error())
+	}
+
+	mailingToken := os.Getenv("MAILING_TOKEN")
+	req.Header.Set("Authorization", mailingToken)
+	req.Header.Set("Content-Type", "application/json")
+
+	resp, err := client.Do(req)
+
+	if err != nil {
```
asynchroza, line 199: Same here, all of this is going to gravely fail.

IvanObreshkov: Added return

packages/api/controllers/hackathon_register_controller.go
IvanObreshkov: Added return

packages/api/controllers/hackathon_register_controller.go
IvanObreshkov: Please don't cut my legs

Issue #176, opened 2023-02-03
HackAUBG Registration form 


SuelaCark:
This issue requires both Frontend and Backend collaboration.
Frontend will deal with the design implementation of the form and checking if the email, password and other info are filled and correct.
Backend will deal with collecting and storing this information and sending "Successfully registered" emails to applicants.

To check progress and work on this issue please checkout branch: "#176-Registration-form"

For a reference on registration form validation, please checkout this video: https://youtu.be/tIdNeoHniEY

PR #362, opened 2023-03-05
#176 Registration form for HackAUBG

Aleks1908:


packages/web/src/components/spa/HackAUBG/RegistrationForm/RegistrationForm.jsx
```diff
@@ -0,0 +1,731 @@
...
+                // console.log('New registartion was added');
+            })
+            .catch((err) => {
+                console.log(err);
+                alert(err['response']['data']['data']['data']);
+            });
+    };
+    const convertToBool = (data) => {
+        data.prevhackaubgparticipation =
+            data.prevhackaubgparticipation == 'False' ? false : true;
```
asynchroza, line 42: instead of doing this, come up with a function which does something similar: 

```javascript

parseVars(data){
    Object.entries(data).forEach(field => {
      if (field[1] == "True" || field[1] == "False") {
             data[field[0]] = field[1] == "True"? true : false
        }
    }

  return data
}
```

then call this function in `onSubmit`:

```javascript

const onSubmit = (data) => {
         setLoadingAnimation(true);
         data = parseVars(data);
         makeRegistration(data);
```

be advised: this is from the top of my hand, there is the possibility something is wrong in this code

packages/web/src/components/spa/HackAUBG/RegistrationForm/RegistrationForm.jsx
```diff
@@ -0,0 +1,731 @@
...
+
+        // axios
+        //     .get()
+        //     .then(() => {})
+        //     .catch(() => {});
+
+        // request failed
+        setTimeout(() => {
+            setLoadingAnimation(true);
+            setError(
```
asynchroza, line 76: Add these to the catch (failure) clause of the axios request

packages/web/src/components/spa/HackAUBG/RegistrationForm/RegistrationForm.jsx
```diff
@@ -0,0 +1,731 @@
...
+
+    const onSubmit = (data) => {
+        setLoadingAnimation(true);
+        convertToBool(data);
+        makeRegistration(data);
+        // send request to backend
+        // if res is good
+        // block submit button, notify that registration is successful
+        setTimeout(() => {
+            setLoadingAnimation(false);
```
asynchroza, line 61: Add this to the then (success) clause of the axios request

packages/web/src/components/spa/HackAUBG/RegistrationForm/RegistrationForm.jsx
```diff
@@ -0,0 +1,731 @@
...
+    const [apiError, setApiError] = useState(false);
+
+    const makeRegistration = (data) => {
+        axios({
+            method: 'post',
+            url: url + '/api/hackathon/register',
+            headers: {
+                'BEARER-TOKEN': localStorage.getItem('auth_token')
+            },
+            data: { ...data }
```
asynchroza, line 28: you don't need the three dots here, no need to unpack, just pass data without the brackets

packages/web/src/components/spa/HackAUBG/RegistrationForm/RegistrationForm.jsx
```diff
@@ -0,0 +1,731 @@
...
+        setError,
+        formState: { errors } // eslint-disable-line
+    } = useForm({ mode: 'all' });
+
+    const [loadingAnimation, setLoadingAnimation] = useState(false);
+    const [submitPressed, setSubmitPressed] = useState(false);
+    const [submitButtonValue, setSubmitButtonValue] = useState('register'); // eslint-disable-line
+    const [apiError, setApiError] = useState(false);
+
+    const makeRegistration = (data) => {
```
asynchroza, line 21: rename this to registerMember

asynchroza:
@Aleks1908 only messages we should display from the backend is if `max capacity reached` (participants, teams) and `email already present in DB`

Aleks1908:
> @Aleks1908 only messages we should display from the backend is if `max capacity reached` (participants, teams) and `email already present in DB`

So messages like: Partcipant added to particapnts without team collection. Are only for us and should not be desplayed

asynchroza:
> > @Aleks1908 only messages we should display from the backend is if `max capacity reached` (participants, teams) and `email already present in DB`
> 
> So messages like: Partcipant added to particapnts without team collection. Are only for us and should not be desplayed

yes, indeed. If the status code is < 300, we should display successfully registered and block the button. If status code is >= 500 we should display a message which says "Something went wrong, please try again later and if it doesn't work please contact thehub-aubg@gmail.com or shoot us a message in Facebook".

packages/web/src/components/spa/HackAUBG/LandingAnimation/LandingAnimation.jsx
```diff
@@ -57,7 +57,14 @@ export const MatrixWindow = () => {
                     HACKAUBG 5.0
                 </h1>
                 <p>31st March - 2nd April, AUBG - Blagoevgrad</p>
-                <button className="Reg-btn">Go to Registration</button>
+                <button
+                    className="Reg-btn"
+                    onClick={() => {
```
asynchroza, line 62: We don't do this anymore, use the `navigateTo` function from Global

packages/web/src/components/spa/HackAUBG/RegistrationForm/RegistrationForm.jsx
```diff
@@ -0,0 +1,732 @@
...
+    } = useForm({ mode: 'all' });
+
+    const [loadingAnimation, setLoadingAnimation] = useState(false);
+    const [submitPressed, setSubmitPressed] = useState(false);
+    const [submitButtonValue, setSubmitButtonValue] = useState('Register'); // eslint-disable-line
+    const [apiError, setApiError] = useState(false);
+    const [buttonMessage, setButtonMessage] = useState('');
+    const [disableTeamNameField, setDisableTeamNameField] = useState(true);
+
+    // ? What are these used for?
```
asynchroza, line 23: Remove comments

packages/web/src/components/spa/HackAUBG/RegistrationForm/RegistrationForm.jsx
```diff
@@ -0,0 +1,732 @@
...
+            if (field[1] === 'True' || field[1] === 'False') {
+                data[field[0]] = field[1] === 'True' ? true : false;
+            }
+        });
+
+        return data;
+    }
+    function getErrorMessage(err) {
+        if (err['response']['status'] >= 500) {
+            setButtonMessage('Something went wrong');
```
asynchroza, line 58: Good, I like it. Consider renaming the function tho, probably something like `setErrorMessage`
Also you should inform the user that they should contact us if they keep experiencing a "Something went error" issue.

packages/web/src/components/spa/HackAUBG/RegistrationForm/RegistrationForm.jsx
```diff
@@ -0,0 +1,732 @@
...
+        } else {
+            setButtonMessage(err['response']['data']['message']);
+        }
+    }
+    function checkTeamname(data) {
```
asynchroza, line 63: Consider adding short comments explaining what each function does.

packages/web/src/components/spa/HackAUBG/RegistrationForm/RegistrationForm.jsx
```diff
@@ -0,0 +1,756 @@
...
+                setSubmitPressed(false);
+                setSubmitButtonValue('Retry');
+                return;
+            } else if (Object.keys(errors).length == 0 && submitPressed) {
+                setButtonState('hackaubg-register-btn disabled');
+                setSubmitButtonValue('Success');
+                return;
+            }
+        }
+        setButtonState('hackaubg-register-btn ');
```
asynchroza, line 111: is the className working as expected with the space at the end?

packages/web/src/components/spa/HackAUBG/RegistrationForm/RegistrationForm.jsx
```diff
@@ -0,0 +1,756 @@
...
+                                        message:
+                                            'No special characters and trailing spaces'
+                                    }
+                                })}
+                            />
+                        </label>
+                        <p className="error-msg">{errors.fullname?.message}</p>
+                    </div>
+                    <div className="send-info">
+                        <label htmlFor="">
```
asynchroza, line 180: What do we use htmlFor for?

packages/web/src/components/spa/HackAUBG/RegistrationForm/RegistrationForm.jsx
Aleks1908: it is working with both with and without it so i will remove it

packages/web/src/components/spa/HackAUBG/RegistrationForm/registration_form.css
```diff
@@ -0,0 +1,209 @@
...
+
+input {
+    text-indent: 2%;
+}
+.invisible {
+    display: none;
+}
+.visible {
+    display: initial;
+}
```
asynchroza, line 148: here as well

packages/web/src/components/spa/HackAUBG/RegistrationForm/registration_form.css
```diff
@@ -0,0 +1,209 @@
...
+
+option:checked {
+    color: #19b700;
+}
+
+input {
+    text-indent: 2%;
+}
+.invisible {
```
asynchroza, line 143: add spaces here

packages/web/src/components/spa/HackAUBG/RegistrationForm/RegistrationForm.jsx
```diff
@@ -0,0 +1,810 @@
...
+                />
+            );
+        }
+    };
+    if (isFormAvailable) {
+        return (
+            <div className="registration-main" id="registration">
+                <h1>Register for HackAUBG 5.0</h1>
+                <form
+                    action=""
```
asynchroza, line 163: what is action?

packages/web/src/components/spa/HackAUBG/RegistrationForm/RegistrationForm.jsx
```diff
@@ -0,0 +1,756 @@
...
+    function checkTeamname(data) {
+        if (data.hasteam == false) {
+            data.teamname = null;
+        }
+        return data;
+    }
+
+    //changes the button color and text depending on what is enetred in the form
+    const checkButtonAvailability = () => {
+        if (apiError == true) {
```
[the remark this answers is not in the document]
asynchroza: I don't see this being updated? Why are the comments resolved?

packages/web/src/components/spa/HackAUBG/RegistrationForm/RegistrationForm.jsx
```diff
@@ -0,0 +1,756 @@
...
+            setButtonState('hackaubg-register-btn error');
+            return;
+        } else if (apiError == false) {
```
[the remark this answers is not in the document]
asynchroza: same here. please if you won't work on something first discuss the reason then resolve the conversation

packages/web/src/components/spa/HackAUBG/RegistrationForm/RegistrationForm.jsx
Aleks1908: probably leftover code

packages/web/src/components/spa/HackAUBG/JourneySection/JourneyMobile/JourneyMobile.jsx
```diff
@@ -12,6 +12,7 @@ export const JourneyMobile = () => {
             </div>
             <div className="journey-mobile-content">
                 <JourneyStep
+                    id="1"
                     title={
                         <h2>
```
asynchroza, line 17: On another note, why the tags are placed here? Shouldn't you pass only the string without the elements (<br>, <s>, <em> tags are exceptions to this) ? Please extract the header tags inside the JourneyStep component

Aleks1908: so i put the id tags there cuz i want ti get the length of the component and before when one is opened it affected all other components

Aleks1908: i will look at it

packages/web/src/components/spa/HackAUBG/RegistrationForm/RegistrationForm.jsx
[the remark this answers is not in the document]
Aleks1908: It is fixed on my end 
image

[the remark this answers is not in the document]
asynchroza: You are still using else if instead of just an else

PR #364, opened 2023-03-05
Fix json keys in edit struct for hackathon team members

### LINKED ISSUE #363: Fix json keys in hackathon team members edit struct

IvanObreshkov:
Fix #363


asynchroza:
Did you test whether editing a member works as expected? We don't have an integration test for this so it would be kind of hard to pinpoint if all is set as expected. If you have tested and everything works as expected, merge the PR.

asynchroza (approved):

IvanObreshkov:
> 

Found huuuuge problem. Cannot update hasTeam value via put request. Working on fixing it now.

IvanObreshkov:
> > 
> 
> Found huuuuge problem. Cannot update hasTeam value via put request. Working on fixing it now.

Fixed

IvanObreshkov:
Fixed key names in all structs, because participants were added to the DB with the wrong names for the fields (the old names of the keys in the JSON part of the struct)

PR #370, opened 2023-03-05
Add cleanup for _tests collection

asynchroza:
Adds a cleanup which drops the test collection and makes sure there are no leftover document entries in the DB after running the tests. This ensures we can create entries with the same data and don't get bad request responses from the API.


IvanObreshkov (approved):
How do you even come up with stuff like this :D

PR #374, opened 2023-03-06
373 width scrolling

SuelaCark:
Removed horizontal scrollbar in HackAUBG


packages/web/src/components/spa/HackAUBG/hack_aubg.css
```diff
@@ -1,5 +1,5 @@
 .hackaubg-container {
-    overflow: scroll;
+    overflow: hidden;
```
asynchroza, line 2: remove the overflow, you don't need it

asynchroza (commented):
Also PRs and commits should have the hashtag when referencing the issues because that's how they get linked to the issue. The idea is only for branch to not feature the hashtag so that it's not hard for people to switch branches

Issue #380, opened 2023-03-07
Missing conditional logic in registration endpoint


asynchroza:
## Brief description:
* When `teams > 15`, we don't return with appropriate message
* Combine the logic of the two functions which check whether the member has already been registered -> you should check both collections at the initial stage and cancel the execution if email is find in either one of them 
* Fix messages in endpoint responses

IvanObreshkov:
Fixed

PR #381, opened 2023-03-07
#380 Fix unhandled case when max number of teams in the hackathon is reached (reg endpoint)

IvanObreshkov:
Add return statement when max # of teams is reached
Move CheckIfNoTeamParticiapntExists at the initial stage


asynchroza (approved):

asynchroza:
Gj

PR #382, opened 2023-03-07
#376 Remove required tag from TeamName field in TeamMember struct 

### LINKED ISSUE #376: Make TeamName field in TeamMember struct not required

IvanObreshkov:
Fix #376

### DIFF
--- a/packages/api/models/hackathon_team_member_model.go
+++ b/packages/api/models/hackathon_team_member_model.go
@@ -6,7 +6,7 @@ type TeamMember struct {
 	ID                             primitive.ObjectID `json:"id" bson:"_id,omitempty"`
 	FullName                       string             `json:"fullname" validate:"required"`
 	HasTeam                        *bool              `json:"hasteam" validate:"required"`
-	TeamName                       string             `json:"teamname" validate:"required"`
+	TeamName                       string             `json:"teamname"`
 	Email                          string             `json:"email" validate:"required"`
 	University                     string             `json:"university" validate:"required"`
 	Age                            int                `json:"age" validate:"required"`


asynchroza (approved):
@IvanObreshkov did you verify if everything is working as expected? If all three cases are handled as expected, feel free to merge this

IvanObreshkov:
> @IvanObreshkov did you verify if everything is working as expected? If all three cases are handled as expected, feel free to merge this

The normal case when the hasTeam is false works as expected (the person is added to the noTeamCollection)
Aleks said in the front-end if the person says he has a team and fills nothing it it si going to notify them that this filed is required. And also when the person clicks No on hasTeam TeamName is going to become inactive.

asynchroza:
> > @IvanObreshkov did you verify if everything is working as expected? If all three cases are handled as expected, feel free to merge this
> 
> @asynchroza The normal case when the hasTeam is false and teamName is "" works as expected (the person is added to the noTeamCollection) Aleks said in the front end if the person says he has a team and fills nothing in teamName it is going to notify them that this filed is required. And also when the person clicks No on hasTeam TeamName is going to become inactive.

It's preferable that you test by not actually passing the key (since this will be the way how the frontend will pass the object to the api), but "" is still considered  a null value so I don't expect that there would be any issue with that. 

I kind of get lost in what you're trying to explain. We handled the issue on the frontend where the field is visible even when `hasTeam` is false. This PR will fix the backend part of the problem. 

CC: @Aleks1908

asynchroza:
> > > > @IvanObreshkov did you verify if everything is working as expected? If all three cases are handled as expected, feel free to merge this
> > > 
> > > 
> > > @asynchroza The normal case when the hasTeam is false and teamName is "" works as expected (the person is added to the noTeamCollection) Aleks said in the front end if the person says he has a team and fills nothing in teamName it is going to notify them that this filed is required. And also when the person clicks No on hasTeam TeamName is going to become inactive.
> > 
> > 
> > It's preferable that you test by not actually passing the key (since this will be the way how the frontend will pass the object to the api), but "" is still considered a null value so I don't expect that there would be any issue with that.
> > I kind of get lost in what you're trying to explain. We handled the issue on the frontend where the field is visible even when `hasTeam` is false. This PR will fix the backend part of the problem.
> > CC: @Aleks1908
> 
> I tested it without providing the teamname key, and the document appears in the DB with the kye as "teamname": "", which I think is the way it should be.

yep, that's not an issue

IvanObreshkov:
> > > > > @IvanObreshkov did you verify if everything is working as expected? If all three cases are handled as expected, feel free to merge this
> > > > 
> > > > 
> > > > @asynchroza The normal case when the hasTeam is false and teamName is "" works as expected (the person is added to the noTeamCollection) Aleks said in the front end if the person says he has a team and fills nothing in teamName it is going to notify them that this filed is required. And also when the person clicks No on hasTeam TeamName is going to become inactive.
> > > 
> > > 
> > > It's preferable that you test by not actually passing the key (since this will be the way how the frontend will pass the object to the api), but "" is still considered a null value so I don't expect that there would be any issue with that.
> > > I kind of get lost in what you're trying to explain. We handled the issue on the frontend where the field is visible even when `hasTeam` is false. This PR will fix the backend part of the problem.
> > > CC: @Aleks1908
> > 
> > 
> > I tested it without providing the teamname key, and the document appears in the DB with the kye as "teamname": "", which I think is the way it should be.
> 
> yep, that's not an issue

Gud, merging

PR #384, opened 2023-03-07
#383 Fix counting of hackathon participants

### LINKED ISSUE #383: Fix counting of all participants
## Brief description:
Now the counting happens only in one collection, and it should be counting from both (the one for participants with team and the one for those without a team

IvanObreshkov:
Fix #383 
There are currently more than 120 documents in the DB. 
image

### DIFF
--- a/packages/api/controllers/hackathon_register_controller.go
+++ b/packages/api/controllers/hackathon_register_controller.go
@@ -51,11 +51,10 @@ func RegisterTeamMember(c *fiber.Ctx) error {
 		ShareInfoWithSponsors:          member.ShareInfoWithSponsors,
 		WantJobOffers:                  member.WantJobOffers}
 
-	numberOfParticipantsWithTeam, _ := GetNumberOfHackathonParticipants(ctx)
-	numberOfParticipantsWithoutTeam, _ := GetNumberOfNoTeamParticipants(ctx)
+	numberOfTotalParticipants := GetTotalNumberOfParticipants()
 	numberOfTeams, _ := GetNumberOfHackathonTeams(ctx)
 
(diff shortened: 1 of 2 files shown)


packages/api/controllers/hackathon_register_controller.go
```diff
@@ -51,10 +51,11 @@ func RegisterTeamMember(c *fiber.Ctx) error {
 		ShareInfoWithSponsors:          member.ShareInfoWithSponsors,
 		WantJobOffers:                  member.WantJobOffers}
 
-	numberOfMembers, _ := GetNumberOfHackathonParticipants(ctx)
+	numberOfParticipantsWithTeam, _ := GetNumberOfHackathonParticipants(ctx)
+	numberOfParticipantsWithoutTeam, _ := GetNumberOfNoTeamParticipants(ctx)
 	numberOfTeams, _ := GetNumberOfHackathonTeams(ctx)
 
-	if numberOfMembers >= 90 {
+	if numberOfParticipantsWithTeam+numberOfParticipantsWithoutTeam >= 90 {
```
asynchroza, line 58: Why don't you just extract this in one function? Either way we would always need to traverse through the two collections. 

e.g. 
```
number_of_participants = GetTotalParticipants()
```

IvanObreshkov: ok

IvanObreshkov: After the registration and probably the hackathon I would try to do a refactoring of the whole backend for the registration.

asynchroza (approved):
merge if all works as expected

Issue #130, opened 2022-12-18
Add public read access to S3 objects by default


asynchroza:
Images should be given public read access by default. Otherwise we would need to do it each time we upload an object

asynchroza:
```json
{
    "Version": "2012-10-17",
    "Id": "Policy1671453606289",
    "Statement": [
        {
            "Sid": "Stmt1671453598171",
            "Effect": "Allow",
            "Principal": "*",
            "Action": "s3:GetObject",
            "Resource": "arn:aws:s3:::hubarskibucket/*"
        }
    ]
}
```

PR #386, opened 2023-03-07
#356 fix jobs section button spacing

Aleks1908:
# Reference the ISSUE you are working on in the name of the PR
e.g. #129 Added webhooks for triggering something (#130)


SuelaCark (changes requested):
Everything looks good and working as expected. However, I would like to remove a little the space between the navbar and the cards.

image

PR #387, opened 2023-03-08
#385 Create endpoint for checking if reg is avaliable

### LINKED ISSUE #385: Endpoint for checking whether Hackathon registration is available
## Brief description:
Write an endpoint which returns either true or false depending on whether the maximum capacity of participants has been reached.

e.g.
GET `/api/hackathon/register/available`

This should be used for hiding the reg form and displaying that the max capacity has been reached @Aleks1908

IvanObreshkov:
Fix #385 
Check if the number of participants is >= 90
if true -> returns 409 status code
if false -> returns 200 status code

### DIFF
--- a/packages/api/controllers/hackathon_register_controller.go
+++ b/packages/api/controllers/hackathon_register_controller.go
@@ -51,8 +51,17 @@ func RegisterTeamMember(c *fiber.Ctx) error {
 		ShareInfoWithSponsors:          member.ShareInfoWithSponsors,
 		WantJobOffers:                  member.WantJobOffers}
 
-	// numberOfTotalParticipants := GetTotalNumberOfParticipants()
 	numberOfTeams, _ := GetNumberOfHackathonTeams(ctx)
+
+	isEligeble, getNubmerErr := IsTotalNumberOfParticipantsEligeble(c)
+
+	if getNubmerErr != nil {
+		return c.Status(http.StatusInternalServerError).JSON(responses.MemberResponse{Status: http.StatusInternalServerError, Message: "error", Data: &fiber.Map{"data": "Couldn't get number of participants, please contact TheHubAUBG"}})
+	}
+
+	if !isEligeble {
+		return c.Status(http.StatusConflict).JSON(responses.MemberResponse{Status: http.StatusConflict, Message: "Max number of Hackathon participants is reached"})
+	}
 
 	if CheckIfTeamMemberExists(newHackathonTeamMember) {
 		return c.Status(http.StatusBadRequest).JSON(responses.MemberResponse{Status: http.StatusBadRequest, Message: "This email is already present in the DB", Data: &fiber.Map{"data": newHackathonTeamMember.Email}})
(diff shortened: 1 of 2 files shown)


packages/api/controllers/hackathon_register_controller.go
```diff
@@ -51,13 +51,9 @@ func RegisterTeamMember(c *fiber.Ctx) error {
 		ShareInfoWithSponsors:          member.ShareInfoWithSponsors,
 		WantJobOffers:                  member.WantJobOffers}
 
-	numberOfTotalParticipants := GetTotalNumberOfParticipants()
+	// numberOfTotalParticipants := GetTotalNumberOfParticipants()
 	numberOfTeams, _ := GetNumberOfHackathonTeams(ctx)
 
-	if numberOfTotalParticipants >= 90 {
```
asynchroza, line 57: You removed this because it's expected they won't be able to hit this endpoint because the form is hidden (e.g. request to `/available` endpoint returned 409 and we've disabled it), right? Well, that's a good thing you paid attention to that, unfortunately, we still need this within the `/register` endpoint. Why? Because these people are supposed to be devs and they might try to register for the hackathon by hitting the endpoint directly. This way we allow them to register even if the cap is reached when using for example Postman, but not our website.

packages/api/controllers/hackathon_register_controller.go
```diff
@@ -164,13 +160,16 @@ func CompareTeamNames(passedTeamName string, storedTeamName string) bool {
 	return FormatTeamName(passedTeamName) == FormatTeamName(storedTeamName)
 }
 
-func GetTotalNumberOfParticipants() int64 {
+func GetTotalNumberOfParticipants(c *fiber.Ctx) error {
```
asynchroza, line 163: You could probably pass the context of the /register endpoint to this one and just check the result

asynchroza (commented):
Good job, needs some refinement

packages/api/controllers/hackathon_register_controller.go
IvanObreshkov: Yes. But I didn't think about protecting us from our own tribe

packages/api/controllers/hackathon_register_controller.go
IvanObreshkov: Ok I will check

packages/api/controllers/hackathon_register_controller.go
```diff
@@ -163,12 +163,23 @@ func CompareTeamNames(passedTeamName string, storedTeamName string) bool {
...
-	numberOfParticipantsWithoutTeam, _ := GetNumberOfNoTeamParticipants(ctx)
+
+	numberOfParticipantsWithTeam, err := GetNumberOfHackathonParticipants(ctx)
+	if err != nil {
+		return c.Status(http.StatusInternalServerError).JSON(responses.MemberResponse{Status: http.StatusInternalServerError, Message: "error", Data: &fiber.Map{"data": "Couldn't get number of participants with a team"}})
+	}
+
+	numberOfParticipantsWithoutTeam, err := GetNumberOfNoTeamParticipants(ctx)
+	if err != nil {
+		return c.Status(http.StatusInternalServerError).JSON(responses.MemberResponse{Status: http.StatusInternalServerError, Message: "error", Data: &fiber.Map{"data": "Couldn't get number of participants without a team"}})
```
asynchroza, line 174: Good

packages/api/controllers/hackathon_register_controller.go
```diff
@@ -51,9 +51,18 @@ func RegisterTeamMember(c *fiber.Ctx) error {
 		ShareInfoWithSponsors:          member.ShareInfoWithSponsors,
 		WantJobOffers:                  member.WantJobOffers}
 
-	// numberOfTotalParticipants := GetTotalNumberOfParticipants()
 	numberOfTeams, _ := GetNumberOfHackathonTeams(ctx)
 
+	isEligeble, getNubmerErr := IsTotalNumberOfParticipantsEligeble(c)
```
IvanObreshkov, line 56: I did it in this way, because when I called the previous function which as you can see returned status codes, but when called here I didn't get them, so I decided to make it with bool and arr. In that way it is going to return the proper status code both here and when CheckRegistrationAvailability is called on GET to `/api/hackathon/register/available`

asynchroza: Second of all, you should be able to pass down the request context as follows:

`/api/hackathon/register` -> `/api/hackathon/register/availability` -> check returned interface

I bet there is a way to check the return value

packages/api/controllers/hackathon_register_controller.go
```diff
@@ -164,13 +166,30 @@ func CompareTeamNames(passedTeamName string, storedTeamName string) bool {
...
-	result := numberOfParticipantsWithTeam + numberOfParticipantsWithoutTeam
-	return result
+	// if err != nil {
+
+	// }
+
+	if numberOfParticipantsWithTeam+numberOfParticipantsWithoutTeam >= 90 {
+		c.SendStatus(400)
+		var err fiber.Error
+		err.Code = 400
```
asynchroza, line 186: Probably swap this for 503

packages/api/controllers/hackathon_register_controller.go
```diff
@@ -164,13 +166,30 @@ func CompareTeamNames(passedTeamName string, storedTeamName string) bool {
 	return FormatTeamName(passedTeamName) == FormatTeamName(storedTeamName)
 }
 
-func GetTotalNumberOfParticipants() int64 {
+func IsRegistrationAvailable(c *fiber.Ctx) error {
 	ctx, cancel := context.WithTimeout(context.Background(), 10*time.Second)
 	defer cancel()
+
 	numberOfParticipantsWithTeam, _ := GetNumberOfHackathonParticipants(ctx)
+	// if err != nil {
```
asynchroza, line 174: Handle exceptions and return error specific messages

packages/api/controllers/hackathon_register_controller.go
```diff
@@ -51,11 +52,12 @@ func RegisterTeamMember(c *fiber.Ctx) error {
...
 
-	numberOfTotalParticipants := GetTotalNumberOfParticipants()
 	numberOfTeams, _ := GetNumberOfHackathonTeams(ctx)
 
-	if numberOfTotalParticipants >= 90 {
-		return c.Status(http.StatusConflict).JSON(responses.MemberResponse{Status: http.StatusConflict, Message: "Max Hackathon participants is reached"})
+	err := IsRegistrationAvailable(c)
+
+	if err != nil {
+		return err
```
asynchroza, line 60: ⚡️⚡️⚡️

asynchroza (approved):
Good stuff

packages/api/controllers/hackathon_register_controller.go
IvanObreshkov: Merging

IvanObreshkov:
> Good stuff

Thank you! It means a lot coming from you!

Issue #278, opened 2023-02-16
Add installation for all necessary plugins [Makefile]


asynchroza:
Current phony installs only `better comments` and `eslint` (if I recall it correctly). 

Updated phony should install the `golang` and `react` plugins we use as well.

PR #388, opened 2023-03-08
#278 Add plugins to Makefile

r-mishev:
Added ***GitLens, Auto Rename Tag, Go, and GoTestExplorer*** extensions to ***install-code-plugins phony***

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -37,7 +37,12 @@ post-wsl:
 install-code-plugins:
 	code --install-extension aaron-bond.better-comments \ 
 	code --install-extension dbaeumer.vscode-eslint \
-	code --install-extension esbenp.prettier-vscode
+	code --install-extension esbenp.prettier-vscode \
+	code --install-extension eamodio.gitlens \
+	code --install-extension formulahendry.auto-rename-tag \
+	code --install-extension golang.Go \
+	code --install-extension premparihar.gotestexplorer
+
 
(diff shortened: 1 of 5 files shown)


Makefile
```diff
@@ -37,7 +37,12 @@ post-wsl:
 install-code-plugins:
 	code --install-extension aaron-bond.better-comments \ 
 	code --install-extension dbaeumer.vscode-eslint \
-	code --install-extension esbenp.prettier-vscode
+	code --install-extension esbenp.prettier-vscode \
+	code --install-extension eamodio.gitlens \
+	code --install-extension formulahendry.auto-rename-tag \
```
asynchroza, line 42: Remove `gitlens`, `auto-rename-tag` and `gotestexplorer`. You may add a recommended plugins section to the README beneath the recommended editor section and include gitlens there. We want the `install-code-plugins` phony to be installing only essential plugins for development. GitLens is a proprietary one so I would prefer that we don't include it in the plugins installation.

asynchroza:
@AUBGTheHUB/nosyncdev any suggestions on what plugins we should add to the default installation? @r-mishev can you check if this is included in the installation scripts and if it's not, please add it to the post installation step.

asynchroza (approved):
Thank you @r-mishev , feel free to merge if there isn't anything you want to add.

Issue #390, opened 2023-03-08
Deleting a team should move all members to the collection of participants without a team


asynchroza:
## Brief description
Delete request for a specific team should move all members from the id list to the collection for participants without a team. 

As we are rushing into things, let @nikolayninov know whether one of you @peti-krl @IvanObreshkov would be able to pull off this ticket. He can manage this part in the frontend but we would need to fix this at first notice after the hackathon. We should leave such an essential part of the registration process as tech-debt.

PR #392, opened 2023-03-09
#390 Move all members from the deleted team to the noTeamParticipants collection

nikolayninov:


packages/api/controllers/hackathon_team_controller.go
```diff
@@ -202,6 +202,27 @@ func DeleteHackathonTeams(c *fiber.Ctx) error {
...
+
+	var tempTeamMember models.TeamMember
+
+	for i := 0; i < len(team.TeamMembers); i++ {
+		key_from_hex, _ := primitive.ObjectIDFromHex(team.TeamMembers[i])
+		_ = teamMembersCollection.FindOne(ctx, bson.M{"_id": key_from_hex}).Decode(&tempTeamMember)
+		teamMembers[i] = tempTeamMember
+	}
+
+	BatchAddTeamMembersToNoParticipants(ctx, teamMembers)
```
asynchroza, line 223: Can you handle exceptions here?

asynchroza (approved):
Please, handle possible exceptions, otherwise it looks good ⚡️⚡️

packages/api/controllers/hackathon_team_controller.go
nikolayninov: Should it return 500 as status code or?

asynchroza: yep

```go
_, err := BatchAddTeamMembersToNoParticipants(ctx, teamMembers)
if err != nil {
    return c.Status(blah, blah, blah, 500, message: "Failed moving members")
}
 
```

asynchroza:
nini, add the hashtag with the number at the beginning of your commits msgs as well.

packages/api/controllers/hackathon_team_controller.go
nikolayninov: Yeah I did that, just needed the green light to commit

asynchroza: don't be afraid to commit and push 😄. We can easily revert back if something is not ok. By pushing I get notified as a codeowner and I can check the code asap if all is good.

Issue #394, opened 2023-03-09
Elements rendering is not fast enough when scrolling animation is invoked for longer than usual lengths


asynchroza:
## Reproduce:
On mobile resolution -> open nav -> click a bottom level anchor

PR #402, opened 2023-03-11
#394 Add body background colour

r-mishev:
Since there is no time and it's the same as the previous website, a custom scrolling wasn't implemented. Rather, a background colour was added to the HubAUBG body, so there's no white boxes popping around.

### DIFF
--- a/packages/web/src/components/spa/HackAUBG/HackAUBG.jsx
+++ b/packages/web/src/components/spa/HackAUBG/HackAUBG.jsx
@@ -20,6 +20,7 @@ export const HackAUBG = () => {
     makeBodyScrollable();
 
     document.body.className = 'hackaubg-container';
+    document.body.style.backgroundColor = '#222222';
 
     const anchorList = [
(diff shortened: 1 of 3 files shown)


packages/web/src/components/spa/HackAUBG/HackAUBG.jsx
```diff
@@ -20,6 +20,7 @@ export const HackAUBG = () => {
     makeBodyScrollable();
 
     document.body.className = 'hackaubg-container';
+    document.body.style.backgroundColor = '#222222';
```
asynchroza, line 23: Since probably this value isn't overwritten when we hop back to the main page, a better approach would be to bind this logic to the function which changes the favicons. 

You could find the function in Global.js. Essentially it's an event listener which checks whether the url contains hackaubg or not. Copy this line of code underneath the conditional statement for the former and set a background Color for the main page as well.

asynchroza (approved):

PR #420, opened 2023-03-15
#419 fixed make gum in Client(prod)

SuelaCark:
Fixed Client menu for Client(prod) and Client(dev)

### DIFF
--- a/cli.sh
+++ b/cli.sh
@@ -53,19 +53,19 @@ if [ $ACTIONS == $START ]; then
     echo -e "What instance do you want to spin up?"
 
     WEB_CLIENT="Client (local)"
-    PROD_CLIENT="Client (dev)"
-    DEV_CLIENT="Client (prod)"
+    DEV_CLIENT="Client (dev)"
+    PROD_CLIENT="Client (prod)"
     LOCAL_API="Run Api"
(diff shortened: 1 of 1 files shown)


cli.sh
```diff
@@ -53,19 +53,19 @@ if [ $ACTIONS == $START ]; then
     echo -e "What instance do you want to spin up?"
 
     WEB_CLIENT="Client (local)"
-    PROD_CLIENT="Client (dev)"
-    DEV_CLIENT="Client (prod)"
+    DEV_CLIENT="Client (dev)"
```
asynchroza, line 56: Nice catch

cli.sh
```diff
@@ -53,19 +53,19 @@ if [ $ACTIONS == $START ]; then
...
+    PROD_CLIENT="Client (prod)"
     LOCAL_API="Run Api"
-    ACTIONS=$(gum choose --limit 1 "$WEB_CLIENT" "$PROD_CLIENT" "$DEV_CLIENT" "$LOCAL_API")
+    ACTIONS=$(gum choose --limit 1 "$WEB_CLIENT" "$DEV_CLIENT" "$PROD_CLIENT" "$LOCAL_API")
```
asynchroza, line 59: We wanted the prod option to be the second one as it is supposed to be the most commonly run. Leave it as you desire, both variants make logic to me.

asynchroza (approved):

PR #449, opened 2023-04-10
#430 Fix security token not being rotated

### LINKED ISSUE #430: [BUG][API] CRON job is not rotating tokens

asynchroza:
Resolves #430


IvanObreshkov (approved):
Is that really the fix? Wow🫨

PR #455, opened 2023-05-04
#454 Fix make gum to work with new repo name

Aleks1908:
Added a conditional statement for `make gum` to work with the new name of the repository


asynchroza (approved):
LGTM and works as expected!

PR #462, opened 2023-08-16
Refactor Golang API

### LINKED ISSUE #18: Clean up API readme - rewrite in markdown
You can use `VSCode` to see the parsed version of the `markdown` file by clicking on the displayed icon (top right corner when editor is open):

image

Refer to the following markdown [documentation](https://www.markdownguide.org/cheat-sheet) 

Make sure you are working on a branch! Refer to the following [wiki](https://github.com/AUBGTheHUB/spa-website-2022/wiki/How-to-work-with-Issues-and-Pull-Requests)

PATH to the file is: `packages/api/README.md`

### LINKED ISSUE #451: [BUG] Hot reloading the API for the first time fails due to incorrect location
## Brief description:
When installing air for the first time, air command fails because it tries to run a config file which is not present at the current location. It happens because we go up two dirs and we don't go back to where the initial config is.

asynchroza:
Removed controllers for managing logic of hackathon teams, hackathon team members, articles and events.
Fixed bug with installing hooks. 

It's expected that some of those controllers will be reimplemented in the new API. 
I left some of the frontend code for managing the hackathon members and teams since it can be reused in the future. 

resolves #451 
resolves #18


packages/api/controllers/utils.go
IvanObreshkov, line 1: good this could be used in many places

IvanObreshkov (approved):
LGTM. But definitely, we need a new API. For this one, if we are keeping the admin panel we have to come up with a better implementation of the auth token.

asynchroza:
> LGTM. But definitely, we need a new API. For this one, if we are keeping the admin panel we have to come up with a better implementation of the auth token.

@IvanObreshkov I don't think we should bother with coming up with a new implementation. This one is sufficient. What we should do is come up with a way to make the two APIs share the token between them. This can be achieved by either creating a new endpoint for verifying the bearer token on the side of the Golang API or update the CRON job to hit an endpoint on the new API which would then load the token in memory. Latter one is more error prone because if the http request fails to update the new API, APIs won't have the same token loaded in memory. Consequentially, half of the admin panel will stop working. Some safety mechanisms can be put in place to prevent such cases but this will be much more demanding on the Golang API than having the new API just verify each request by hitting the Golang API. 

image

asynchroza:
resolves #462 
resolves #18

PR #470, opened 2023-08-23
#98 Drop actions endpoint if stateless

### LINKED ISSUE #98: Cannot access action endpoint when url is pasted in new tab
## Brief description:
When directly accessing `/admin/dashboard/<section>/actions` endpoint (copy pasting the url to another browser/tab), page doesn't get rendered because location [state](https://github.com/AUBGTheHUB/spa-website-2022/blob/3a0e12132f8238d7b4463edc2bcb173842d130f1/packages/web/src/components/admin_page/articles_page/actions_articles.jsx#L11) is null (page is trying to access them). 

image

## How to achieve it:
Check whether location [state](https://github.com/AUBGTheHUB/spa-website-2022/blob/3a0e12132f8238d7b4463edc2bcb173842d130f1/packages/web/src/components/admin_page/articles_page/actions_articles.jsx#L11) is null, if so redirect to current url without the last chained endpoint.  

e.g. `/admin/dashboard/members/actions` -> location.state is null -> `/admin/dashboard/members`

Define this as a utility function which could be passed an argument (location.state). It should be saved in a file in the root of `admin_page`. 

[Check this file for an example on how to export variables/functions](https://github.com/AUBGTheHUB/spa-website-2022/blob/master/packages/web/src/Global.js)

asynchroza:
resolves #98 

## How to verify:

Manually append `/actions` to an endpoint of section you've chosen from the dashboard.
For example `partners` which lives on `/admin/dashboard/partners`
image

Try to access `/admin/dashboard/partners/actions` and you should be returned back to the previous page no matter whether you do it in the same tab or in a new one.

### DIFF
--- a/.github/CODEOWNERS
+++ b/.github/CODEOWNERS
@@ -1,14 +1,12 @@
-.githooks @asynchroza @SuelaCark @nikolayninov
-
 .github/workflows @asynchroza
 
 .vscode @asynchroza
 
 data @asynchroza
 
-packages/web @SuelaCark @Aleks1908 @asynchroza @nikolayninov @DemonHashira
+packages/web @SuelaCark @Aleks1908 @asynchroza @nikolayninov @DemonHashira @r-mishev
 
-packages/api @nikolayninov @IvanObreshkov @asynchroza @peti-krl
+packages/api @nikolayninov @IvanObreshkov @asynchroza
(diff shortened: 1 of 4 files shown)


Aleks1908 (approved):
LGTM

PR #482, opened 2023-08-25
Rework structure to use separate routers

asynchroza:
<s>POC for a new structure which will allow the creation of new APIRouter-s.

How it is achieved:
A second FastAPI app is being created and mounted to the main one in order to have the required proxy prefix `/v2`.

A decorator is being appended to the `bind()` function of the routes classes to bind the paths to the app after their initialisation. 

This way we can still stick to having a higher level of encapsulation and still utilize the additional features provided by the APIRouter class.

Additionally, I added a new abstract base class which is used when creating routes classes in order to ensure an additional level of hand holding when implementing new routes. This way app will throw an exception if abstract method used for binding the routes to the app is not implemented. </s>

F*ck it, we ball.
We go with simpler code. Declare router and routes, import router in `routes.py`, add in `routers` list - done. A couple of abstraction layers removed.

@IvanObreshkov I think this will make it more readable for you.


packages/py-api/py_api/routes/routes.py
```diff
@@ -1,18 +1,24 @@
...
 
 class Routes:
-    @staticmethod
-    def bind(router: APIRouter) -> None:
-        UtilityRoutes.bind(router)
-        UploaderRoutes.bind(router)
-        UrlShortenerRoutes.bind(router)
+    def bind(app: FastAPI) -> None:
+        for router in routers:
+            app.include_router(router)
```
IvanObreshkov, line 24: Nice. We actually could get by without the class Routes and just have a bind function to which we pass the instance of our app (otherwise if we import it directly we would get circular import). But I prefer the OOP way, It keeps things organized and easier to manage.

packages/py-api/py_api/routes/url_shortener_routes.py
```diff
@@ -1,21 +1,21 @@
...
-    @staticmethod
-    def bind(router: APIRouter) -> None:
-        @router.get("/shortener")
-        def get_shortened_urls() -> Dict[str, Any]:
-            return c.fetch_shortened_urls()
 
-        @router.delete("/shortener/{endpoint}")
-        def delete_shortened_url(endpoint: str) -> Dict[str, Any]:
-            return c.delete_shortened_url(endpoint)
+def get_shortened_urls() -> Dict[str, Any]:
```
IvanObreshkov, line 10: I believe you forgot the router.get("")

packages/py-api/py_api/routes/utility_routes.py
```diff
@@ -1,28 +1,28 @@
...
+    return {
+        "status": "healthy",
+        "currentMemoryUsage": f"{get_current_memory_usage_in_mbs()} MB",
+    }
 
-            return {
-                "paths": routes,
-            }
+
+@router.get('/routes')
```
IvanObreshkov, line 17: Nice touch. I see that some of the default routes are listed as well 
image
Do we need them?

IvanObreshkov (changes requested):
Overall, I really like the new concept. It is simpler to understand, at least for me, compared to the previous one. Here we have a main app to which we `bind` or include in all of our routes with the `include_router()` func as shown in the docs of FastAPI. In the previous one, we had like a main route, to which we bounded/appended the other routes via custom implementation. It is basically the same when you look at it from a practical standpoint but could be a bit confusing for someone who reads the documentation of FastAPI and then looks at our code for the first time.

packages/py-api/py_api/routes/routes.py
asynchroza: Yes, we can do that but it won't provide any more efficiency over this one. I guess this solely depends on personal preference. I would prefer to have the bind method encapsulated in a class rather than having it fly around. I bet you've had experienced at least once importing the "right" function from an incorrect module.

packages/py-api/py_api/routes/utility_routes.py
asynchroza: Do you mean to ask whether we need the endpoint or the docs/openapi routes? If we want to exclude the routes from the list we would need to add explicit filtering. I think it's a good idea to keep them as reminders. I didn't know that they exist before writing the endpoint.

packages/py-api/py_api/routes/url_shortener_routes.py
asynchroza: Good catch! Thank you!

asynchroza:
@IvanObreshkov ready for another round of reviews. 
By the way, it's better if you avoid requesting changes on the pull request. This is because doing so would necessitate your approval for the modifications, even if someone else has already reviewed them.

packages/py-api/py_api/routes/routes.py
IvanObreshkov: Yes, totally agree.

packages/py-api/py_api/routes/utility_routes.py
IvanObreshkov: We are keeping them.

IvanObreshkov (approved):
To the next one!

PR #486, opened 2023-08-27
#436 Add admin panel section for url shortener

### LINKED ISSUE #436: [URL SHORTENER] Add a new admin panel section for generating urls
## Brief description:
Add an admin panel section where a user would be able to generate a short url

JSON PAYLOAD: 
```json
{
    "url": <url-to-which-endpoint-is-pointing>,
    "endpoint": <endpoint-user-wants>
}

{
    "url": "https://github.com/AUBGTheHUB/spa-website-2022/issues/436",
    "endpoint": "example"  
}

```
on success, save the following url: https://dev.thehub-aubg.com/s/<endpoint> in the user's clipboard (display notification)
otherwise, inform of fail

Pass token in Authorization header the same way we do in S3 Management

test for example: https://dev.thehub-aubg.com/s/example

asynchroza:
Added a new admin panel section for the URL shortener. 
Similarly to what https://bit.ly is, we have our own shortener. 
For example try accessing https://thehub-aubg.com/s/mono - it will lead you to our repo.

## How to verify:
* Try to create a new shortened url - with and without a proper URL
* Try to update a shortened url - with and without a proper URL
* Try to remove a shortened url

resolves #436 

image
image


Aleks1908 (approved):
LGTM

PR #490, opened 2023-08-28
Fix source issue

Aleks1908:
Fix install_osx.sh issue


IvanObreshkov (approved):
LGTM

mbozhilov-qb:
For safe keeping reasons, I'm leaving this comment here:

@Aleks1908 was having a really strange bug happen to him where none of the node version manager configuration scripts will get installed in the config folder which is being used by the post installation macos script (and the shell as well - `NVM_DIR`). The remedy involved sourcing the installation script from the brew directory during configuration. Subsequently, the process began functioning as intended without requiring extra shell sourcing upon startup, it's just a one time thing which has to be done during the initial setup.

asynchroza:
For safe keeping reasons, I'm leaving this comment here:

@Aleks1908 was having a really strange bug happen to him where none of the node version manager configuration scripts will get installed in the config folder which is being used by the post installation macos script (and the shell as well - NVM_DIR). The remedy involved sourcing the installation script from the brew directory during configuration. Subsequently, the process began functioning as intended without requiring extra shell sourcing upon startup, it's just a one time thing which has to be done during the initial setup.

PR #493, opened 2023-08-28
#491 Remove the need for verification requests to Golang API when running locally

### LINKED ISSUE #491: [FEATURE] Remove the need for a verification request to Go API when running locally
## Brief description:
When running locally, the python API makes a request to the Golang API before continuing with its regular handling of requests. This particular request is used for verifying the caller by passing the BEARER-TOKEN header to the `/api/verify` endpoint.

In order to avoid having the two APIs running at the same time consuming computer resources, we may bypass the authentication mentioned above when we're running in the "OFFLINE" environment.

## How to achieve it:
Check what environment is running and run the authentication using an [offline token](https://github.com/AUBGTheHUB/monolith/blob/49e57cb9b74e024d06efadaa3938be71ef5e8adc/packages/api/configs/env.go#L48). Important to note is that we don't really want to skip the whole authentication logic, we want to emulate production workflow as much as possible. Therefore, you need to load `IS_OFFLINE` environment variable and check whether the bearer token is equal to `OFFLINE_TOKEN` since this is the default value for the token when running in the offline environment.

asynchroza:
Bypass authentication request to Golang API by comparing received BEARER-TOKEN to OFFLINE_TOKEN when running the Python API locally.

resolves #491 

## How to verify:

- [ ] Add a logger in the conditional statement for checking the offline environment in the authentication middleware
- [ ] Check if log message gets displayed and you can run all the requests that you need when passing the correct BEARER-TOKEN

### DIFF
--- a/packages/py-api/py_api/environment.py
+++ b/packages/py-api/py_api/environment.py
@@ -8,3 +8,5 @@ load_dotenv()
 MONGO_URI = getenv("MONGOURI", "")
 
 IS_OFFLINE = eval_bool(getenv("IS_OFFLINE", False))
+
+OFFLINE_TOKEN = "OFFLINE_TOKEN"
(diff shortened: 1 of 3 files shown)


IvanObreshkov (approved):
Good job.

PR #492, opened 2023-08-28
#484 Endpoint for downloading logfiles locally

IvanObreshkov:
Created an endpoint `GET /logs` for downloading the `logfile.log ` from our prod server when deployed.

Resolves #484   

## How to verify:
- [ ] Change the `IS_OFFLINE` value in your .env file to `false`
- [ ] Go to the `auth.py` middleware and add the endpoint to the  BYPASSED_ENDPOINTS dict:
image

- [ ] Open your browser and head to [/v2/logs](http://localhost:6969/v2/logs).

The file should appear in the folder where your Downloads are stored by default. If no text or page appears on your screen this is completely normal as we are returning a `FileResponse`. If the file is empty this is normal as well.
I encourage you to come up with other edge cases, if possible, that are not handled in my code.

PR #497, opened 2023-08-28
#496 Encapsulate contants and add @classmethod where needed

### LINKED ISSUE #496: [FEATURE] Encapsulate constants in classes
## Brief description:
The constants should be encapsulated in the classes they are used in. Also, @classmethod or @staticmethod decorators should be added to all methods defined in classes that do not have them, and we do not require an instance to call those methods.
Examples:
image
image

IvanObreshkov:
* Encapsulated all constants in their representative classes and made them private. Also added @classmethod  where needed.
* Swaped print for a logger in mailer.py

Resolves #496 

## How to verify:
Review my code and test the mailer service when an exception is raised.

### DIFF
--- a/packages/py-api/py_api/controllers/logs_controller.py
+++ b/packages/py-api/py_api/controllers/logs_controller.py
@@ -7,7 +7,7 @@ from py_api.environment import IS_OFFLINE
 
 
 class LogsController:
-    LOGFILE_PATH: Final = f"{Path(__file__).parent.resolve().parent}/shared/logfile.log"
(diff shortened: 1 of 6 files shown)


asynchroza:
A couple of small nits and it should be good to go! Nice job on setting up the class members as private.

IvanObreshkov:
> A couple of small nits and it should be good to go! Nice job on setting up the class members as private.

Should be good to go.

asynchroza (approved):
Good work on this! I don't know how this was working for you but this [type](https://github.com/AUBGTheHUB/monolith/pull/492#discussion_r1307346083) was breaking the API for me. The issue is not MyPy actually but the FastAPI's underlying type inference.

PR #499, opened 2023-08-28
#498 Add basic request analytics

### LINKED ISSUE #498: [FEATURE] Implement Analytics
## Brief description:
Add middleware for tracking number of requests and where they're being made from - country, city.
## How to achieve it:
Figure it out

asynchroza:
Created a middleware for running a background task for updating a document in the database representing the total count of requests made to the server.

The document looks as follows:
image

How to verify:

- [x] Set IS_PROD environment variable to true
- [x] Make a request
- [x] Check document in analytics collection - total_requests should be updated +1 and in locations you should be able to find your city with an updated request count as well

resolves #498

### DIFF
--- a/packages/py-api/py_api/environment.py
+++ b/packages/py-api/py_api/environment.py
@@ -8,5 +8,6 @@ load_dotenv()
 MONGO_URI = getenv("MONGOURI", "")
 
 IS_OFFLINE = eval_bool(getenv("IS_OFFLINE", False))
+IS_PROD = eval_bool(getenv("IS_PROD", False))
 
 OFFLINE_TOKEN = "OFFLINE_TOKEN"
--- a/packages/py-api/py_api/middleware/analytics.py
+++ b/packages/py-api/py_api/middleware/analytics.py
@@ -29,7 +29,9 @@ class AnalyticsMiddleware:
     async def update_entries(cls, request: Request) -> None:
         projection = {"_id": 0}
         analytics = cls.an_col.find_one({}, projection)
-        country, city = cls.get_country_and_city_from_ip(request.client)
+
+        if request.client:
+            country, city = cls.get_country_and_city_from_ip(request.client)
 
         analytics = cls.update_analytics(analytics, country, city)
 
@@ -43,7 +45,7 @@ class AnalyticsMiddleware:
         response = get(
             "https://geolocation-db.com/json/{ip_address}&position=false",
         ).json()
-        return (response["country_name"], response["city"])
+        return response["country_name"], response["city"]
 
     def create_update_operation(locations: Dict[str, Any]) -> Dict[str, Dict[str, Any]]:
         update_operation: Dict[str, Dict[str, Any]] = {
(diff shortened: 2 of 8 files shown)


packages/py-api/py_api/middleware/analytics.py
```diff
@@ -0,0 +1,77 @@
...
+        status = cls.an_col.update_one({}, update_operation, upsert=True)
+
+        if status.modified_count != 1 and not status.acknowledged:
+            logger.error("Couldn't update analytics")
+
+    def get_country_and_city_from_ip(ip_address: str) -> Tuple[str, str]:
+        response = get(
+            "https://geolocation-db.com/json/{ip_address}&position=false",
+        ).json()
+        return (response["country_name"], response["city"])
```
IvanObreshkov, line 46: Remove redundant parentheses

packages/py-api/py_api/environment.py
```diff
@@ -8,5 +8,6 @@
 MONGO_URI = getenv("MONGOURI", "")
 
 IS_OFFLINE = eval_bool(getenv("IS_OFFLINE", False))
+IS_PROD = eval_bool(getenv("IS_PROD", False))
```
IvanObreshkov, line 11: Is this because we are going to have a dev environment as well? I mean  IS_OFFLINE env var is Flase, aren't we in PROD?

packages/py-api/py_api/middleware/analytics.py
```diff
@@ -0,0 +1,77 @@
...
+                    self.update_entries, request,
+                )
+
+            return response
+
+    @classmethod
+    async def update_entries(cls, request: Request) -> None:
+        projection = {"_id": 0}
+        analytics = cls.an_col.find_one({}, projection)
+        country, city = cls.get_country_and_city_from_ip(request.client)
```
IvanObreshkov, line 32: image

packages/py-api/py_api/middleware/analytics.py
```diff
@@ -0,0 +1,77 @@
...
+
+        analytics = cls.update_analytics(analytics, country, city)
+
+        update_operation = cls.create_update_operation(analytics)
+        status = cls.an_col.update_one({}, update_operation, upsert=True)
+
+        if status.modified_count != 1 and not status.acknowledged:
+            logger.error("Couldn't update analytics")
+
+    def get_country_and_city_from_ip(ip_address: str) -> Tuple[str, str]:
```
IvanObreshkov, line 42: This should be @staticmethod

packages/py-api/py_api/middleware/auth.py
```diff
@@ -22,17 +22,15 @@ class AuthMiddleware:
         "/health": ["GET"], "/routes": ["GET"], "/fswitches": ["GET"],
     }
 
-    @classmethod
-    def bind(cls, app: FastAPI) -> None:
-
+    def __init__(self, app: FastAPI) -> None:
```
IvanObreshkov, line 25: Why did you decide to have a constructor? Do we need an instance?

packages/py-api/py_api/environment.py
asynchroza: Correct. `IS_OFFLINE` won't particularly mean that we're on production.

packages/py-api/py_api/middleware/auth.py
asynchroza: So that it matches the approach we use for routes. I decided that only the binding class should have a method bind.

IvanObreshkov: Nice one

IvanObreshkov (approved):
Nice job with clean code as always! Bravo!

Issue #495, opened 2023-08-28
[FEATURE] Exception handler is not logging any information for the request being made


asynchroza:
## Brief description:
As of now, the exception handler will display full traceback but this is not enough when debugging failing requests. 

We need:
- [x] Full url -> this will tell us whether there are any query or path parameters present in the request
- [x] Method type

Additional:
Extract the logic from the health routes into controllers

IvanObreshkov:
@Aleks1908 You said you wanted to try something from the backend. I believe you can tackle this issue easily. Please try without ChatGPT first. 😃

asynchroza:
Pasting this here if it ever comes to us needing the exception handling decorator again - https://github.com/AUBGTheHUB/monolith/pull/502

PR #502, opened 2023-08-30
POC: Exception handler to read body of requests

asynchroza:
Decorator for binding the proper body coroutine to the exception as an attribute.

In exception handler, check if attribute exists and parse the body -> this check can be removed once all controllers are using the decorator if requests without bodies resolve to empty dicts. 

Testing endpoint is `/health` -> main functionality should be extracted in a controller so that we can properly append the decorator.

- [ ] Decorator should log a warning message if arg[0] is not of type request. -> route handlers should always pass the request object to the controllers.

### DIFF
--- a/packages/py-api/py_api/controllers/url_shortener_controller.py
+++ b/packages/py-api/py_api/controllers/url_shortener_controller.py
@@ -3,6 +3,7 @@ from typing import Dict, List
 from fastapi.responses import JSONResponse
 from py_api.database import su_col
 from py_api.models import ShortenedURL
+from py_api.utilities.decorators import handle_exception
 from py_api.utilities.parsers import has_prohibited_characters
 
 
@@ -26,6 +27,7 @@ class UrlShortenerController:
             return JSONResponse(content={"message": "Endpoint wasn't found!"}, status_code=404)
 
     @classmethod
+    @handle_exception
     def upsert_shortened_url(cls, body: ShortenedURL) -> Dict[str, str] | JSONResponse:
         dumped_body = body.model_dump()
         prohibited_chars = "'\";/:!@#$%\\[]^*()_-+{}=?.,§~`"
--- a/packages/py-api/.pre-commit-config.yaml
+++ b/packages/py-api/.pre-commit-config.yaml
@@ -25,8 +25,8 @@
       hooks:
         - id: add-trailing-comma
 
-    -   repo: https://github.com/pre-commit/mirrors-mypy
-        rev: 'v1.5.1'  # Use the sha / tag you want to point at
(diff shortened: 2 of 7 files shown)


packages/py-api/py_api/controllers/url_shortener_controller.py
```diff
@@ -26,6 +27,7 @@ def delete_shortened_url(cls, endpoint: str) -> Dict[str, str] | JSONResponse:
             return JSONResponse(content={"message": "Endpoint wasn't found!"}, status_code=404)
 
     @classmethod
+    @handle_exception
```
asynchroza, line 30: @Aleks1908 @IvanObreshkov 


This is the new decorator that we need to use to add the body to the exception message received by the end user. Honestly, I believe it will make our codebase more cluttered. I suggest that we remove the feature for displaying bodies. Since our APIs are not publicly accessible, in most debugging scenarios, you can already see the request body because you're the one making the request.

WDYT?

IvanObreshkov: Nice job creating the controllers. Also thank you for making the effort to write the decorator. In my opinion aside from our code base being more clutterd, people might forget to add it to the routes. Also for the Post/Put requests we are going to have a body validation via Pydantic, by creating Models. Now for the environments. Both locally and in prod these messages are useful (the url of the requst and the HTTP method), especially in prod. But wouldn't the Body validation of Pydantic keep us safe and prevent the need for displaying the body of the request in the exception message?

asynchroza: I believe so. Essentially we can be even more strict with our Pydantic requirements in order to prevent the usage of incorrect values. For example, if we have a `verify` property on the body which can take only two values, we should enforce a check for those two types and throw a Pydantic exception if the passed value doesn't match none of the expected values. 

Having the body would've been helpful but I think we don't really need it right now. I prefer keeping our codebase consistent and easy to read for the backend newcomers.

IvanObreshkov: Yes, totally agree with you. If you want you can leave the controllers for the health routes. And remove the decorator in order to merge this PR, ot you could create a separate one. Also @Alex you could open an PR for #495

asynchroza: I suggest that @Aleks1908 creates the controllers in his PR and I close this one out. It will help him understand the high level concept of controllers and routes.

IvanObreshkov: Deal🤝

PR #505, opened 2023-08-31
#487 Set up new deployment services

### LINKED ISSUE #487: [FEATURE] Set up new services for deployment
## Tasks:
- [x] Set up reverse proxy for `/v2` (refer `nginx.conf` in web)
- [x] Add python api Dockerfile (refer Dockerfile in api, do not implement stages for now)
- [x] Load certificates in python api (refer golang api or [hubconf api](https://github.com/AUBGTheHUB/hub-conf-signup/blob/097ad0d8bbb6b0cfd1013ed6a322ee257cd5a02a/packages/api/server.js#L73)) <-- prioritize this now
- [x] Add python api in compose with shared volume for logs (refer `docker-compose.yaml` in root)
- [x] Add health check for python api and url shortener in `manage_app.py` (refer `manage_app.py`)
- [x] (OPTIONAL) Update `manage_app.py` to accept new flags for enabling easier debugging locally
```bash
./manage_app.py --no-health-checks --no-cd --no-renewal
```

 `--no-health-checks` disables all health checks
 `--no-cd` disables checking for newer changes
 `--no-renewal` disables the renewal of certificates, enables developers to use self signed certificates 
- [ ] (OPTIONAL) Set up a sandbox for testing the `manage_app.py`. A container which runs the script within itself, sets up all needed certificates and binds all needed ports for enabling local testing. Project's root is a shared volume.
- Loads the environment from the globally shared `.env`, there won't be any need to overwrite our config files manually.
- It will enable for easier testing of the certs installation/rotation and ssl configurations.

IvanObreshkov:
Resolves #487 
## What I did:
* Set up reverse proxy for `/v2` 
*  Added Python API Dockerfile 
* Load certificates in Python API 
* Added Python API in compose with shared volume for logs 
* Added health check for Python API and url shortener in `manage_app.py`
* Added flags for loading env vars from a file, disabling Discord notifications, disabling checking for new changes in the repo, and disabling the renewing of SSL certificates
* Added certs to questionnaires service and added it to the nginx config

## How to verify:
ENV variables:
* IS_OFFLINE=false
* BEARER_TOKEN="OFFLINE_TOKEN"

- [ ]  Do a code review
- [ ]  Run the script locally with each flag and with all flags combined. To view the flags: `python manage_app.py --help`


docker-compose.yml
```diff
@@ -1,61 +1,78 @@
...
-    ports:
-      - 8000:8000
-    command: /api
-    volumes:
-      - ./data/certs:/app/certs
-    networks:
-      - front-back
-
-# WEB
-  client:
```
asynchroza, line 26: This container is the one hosting the nginx server. Therefore this is the first point of contact when making a request. You can't test the reverse-proxy without having this up and running.

docker-compose.yml
```diff
@@ -1,61 +1,78 @@
...
-    build: 
-      context: ./packages/web
-      args:
-        REACT_APP_OBJ_UPLOADER_URL: ${REACT_APP_OBJ_UPLOADER_URL}
```
asynchroza, line 30: These are no longer needed.

docker-compose.yml
```diff
@@ -1,61 +1,78 @@
...
     ports:
-      - 80:80
-      - 443:443
-    command: nginx -g 'daemon off';
-    volumes: 
-      - ./data/certs:/etc/nginx/certs
+      - 6969:6969
+    # command: /py-api
+    # volumes:
+    #   - spa-website-22/packages/py-api/pyapi/shared
```
asynchroza, line 18: I suppose you know that the volume here is not correctly set up and that's why it's commented out.

docker-compose.yml
```diff
@@ -1,61 +1,78 @@
...
       - ".env"
     expose:
-      - 443
+      - 6969
     ports:
-      - 80:80
-      - 443:443
-    command: nginx -g 'daemon off';
-    volumes: 
-      - ./data/certs:/etc/nginx/certs
```
asynchroza, line 42: Follow this as an example how to set volumes. Left is relative local location, right is location in container.

asynchroza: Additionally, since this story includes some work with ssl certificates. In the root of the project there is a directory called /data/certs where we store the certificates. You will probably need to create some self-signed certificates in order to test out whether https works as expected.

packages/py-api/Dockerfile
```diff
@@ -0,0 +1,26 @@
+FROM python:3.11 as build
+
+# For some reason when installing poetry with curl, after that poetry is not found
```
asynchroza, line 3: Because you need to add it to the PATH

asynchroza: but if pip yields the same result, there is no need to play around with using curl

packages/web/nginx.conf
```diff
@@ -34,5 +34,10 @@ server{
     proxy_set_header "Connection" "";
     proxy_pass http://shortener:8001;
   }
-}
 
+  location /v2 {
+    proxy_http_version 1.1;
+    proxy_set_header   "Connection" "";
+    proxy_pass https://localhost:6969;
```
asynchroza, line 41: https://localhost:6969; -> https://<name-of-container>:6969;

packages/py-api/py_api/main.py
```diff
@@ -28,7 +28,7 @@
 def start() -> None:
     run(
         "py_api.main:main_app", host="0.0.0.0", port=6969,
-        reload=True, log_config=get_log_config(),
+        reload=True, log_config=get_log_config(), **SSL_FILES if not IS_OFFLINE else {},
```
asynchroza, line 31: reloading should also be disabled for deployed instances

asynchroza:
Readme:
- [ ] How to run services using docker-compose. 
* environment variables
* certificates
* how to make a request, high level overview of how the proxy works
- [ ] How to run the management script locally for testing.
* arguments
* explain functionality


In a separate issue we should cover the creation of a readme for deployments.

asynchroza:
The management script requires some particular variables to be loaded into the environment in order for it to be able to run. Let's add them to the sample env file.

./manage_app.py --load-env

If this flag is present, use dotenv to load environment variables.

packages/py-api/py_api/main.py
IvanObreshkov: Done

IvanObreshkov:
> @IvanObreshkov any updates here?

Working on it now

manage_app.py
```diff
@@ -41,6 +34,29 @@ class bcolors:
     CEND = '\033[0m'
 
 
+args_parser = ArgumentParser(description="CLI args for the script")
+
+args_parser.add_argument(
+    '--load_env', action='store_true',
```
asynchroza, line 40: Arguments are written with dashes instead of underscores. `--load-env`

asynchroza: I see what you're doing with the underscores. Having the arguments as readable class members. Nevertheless, I think it's better if we stick to the proper naming convention.

manage_app.py
```diff
@@ -107,85 +152,123 @@ def start_docker_compose():
...
+        ],
+        capture_output=True,
+    )
 
     errors["BUILD"] = build_err.stderr.decode('utf-8')
 
     print(bcolors.RED_IN + "BUILD FAILED" + bcolors.CEND)
 
-    msg['Subject'] = f'{ENV}:SPA BUILD FAILED'
+    msg['Subject'] = f'{os.getenv("DOCK_ENV")}:SPA BUILD FAILED'
```
asynchroza, line 258: Why?

manage_app.py
```diff
@@ -264,8 +353,30 @@ def check_service_up(url: str, service: str, discord=False):
...
+        except Exception as e:
+            return handle_exception(msg, req_method, url, service, e, discord)
+        if web_request.status_code != 200:
+            return handle_status_code_exception(
+                msg, req_method, url, service, web_request.status_code, discord,
+            )
+    elif service == "URL-SHORTENER":
 
+        try:
+            # TODO: Add post to Golang API to get token
```
asynchroza, line 372: Why do you need another request? Shortener doesn't have authentication.

manage_app.py
```diff
@@ -130,28 +193,46 @@ def get_commit_url():
...
+            os.getenv("HUB_PY_API_URL"), "PY-API", False,
+        )
+
+        # URL-SHORTENER
+        get_url_shortener = check_service_up(
+            os.getenv("HUB_URL_SHORTENER"), "URL-SHORTENER", False,
+        )
 
         print()
         if (get_web == 200 and get_api == 400):
```
asynchroza, line 216: Should check the status codes of the other services as well

IvanObreshkov: Done

asynchroza:
No description is provided of made changes @IvanObreshkov

packages/web/nginx.conf
```diff
@@ -23,16 +23,16 @@ server{
     proxy_pass https://api:8000;
   }
 
-  location = /s {
+  location /s/ {
```
asynchroza, line 26: I don't see reverse proxy to questionnaire + certs

IvanObreshkov:
> No description is provided of made changes @IvanObreshkov

Sorry mate, one moment

asynchroza (commented):
Otherwise, from a quick look it looks good

IvanObreshkov:
@asynchroza I added the certs to the vite config and the questionnaire service to the nginx.conf, but I am getting a 502 bad gateway. A further examination is needed

mbozhilov-qb:
> @asynchroza I added the certs to the vite config and the questionnaire service to the nginx.conf, but I am getting a 502 bad gateway. A further examination is needed

There a couple of things you have incorrectly set up. Here's a branch out with an example express server https://github.com/AUBGTheHUB/monolith/tree/487-branch-out-poc-server

manage_app.py
```diff
@@ -1,33 +1,61 @@
...
+import requests
 import schedule
+from dotenv import load_dotenv
 
 BUILD_RUNNING = threading.Event()
 CURRENTLY_BUILDING = threading.Event()
 BUILD_FAILED = threading.Event()
 lock = threading.Lock()
 BUILD_TRY = 0
+logger = getLogger("manage-app-logger")
```
asynchroza, line 24: Where are you using this logger? If you're going to be using a logger, swap all print statements for the logger. Otherwise, remove the logger. I prefer the latter.

manage_app.py
```diff
@@ -139,19 +206,37 @@ def get_commit_url():
...
+
+        # URL-SHORTENER
+        get_url_shortener = check_service_up(
+            SHORTENER, "URL-SHORTENER", False,
+        )
+
         print()
-        if (get_web == 200 and get_api == 400):
-            print(bcolors.OKGREEN + f"{ENV} BUILD SUCCESSFUL" + bcolors.CEND)
+        if get_web == 200 and get_api == 400 and get_py_api == 200 and get_url_shortener == 200:
```
asynchroza, line 220: a check for questionnaire is missing

asynchroza:
I'm testing the script on the dev environment and it seems that loading the environment variables into the bashrc file and not having them as an env file causes docker-compose to fail. Therefore, fully migrating to using dotenv might be the better approach here instead of declaring environment variables both in the env file and the rc one.

IvanObreshkov:
> I'm testing the script on the dev environment and it seems that loading the environment variables into the bashrc file and not having them as an env file causes docker-compose to fail. Therefore, fully migrating to using dotenv might be the better approach here instead of declaring environment variables both in the env file and the rc one.
> 
> @IvanObreshkov 

Ok I will change it

manage_app.py
```diff
@@ -236,35 +333,70 @@ def check_service_up(url: str, service: str, discord=False):
...
+        ) + bcolors.CEND,
+    )
 
     req_method = ""
 
     if service == "WEB":
         try:
-            web_request = requests.get(url)
+            web_request = requests.get(
+                url, verify=False if ENV != "PROD" else True,
```
mbozhilov-qb, line 348: ENV != "LOCAL"

asynchroza, line 348: * Why the ternary?
* `verify=ENV != "LOCAL"`
* All deployed environments have certificates and we want to make verified requests to check their status

manage_app.py
```diff
@@ -236,35 +333,70 @@ def check_service_up(url: str, service: str, discord=False):
...
+
+        if web_request.status_code != 200:
+            return handle_status_code_exception(
+                msg, req_method, url, service, web_request.status_code, discord,
+            )
+
+    elif service == "URL-SHORTENER":
+        try:
+            web_request = requests.get(
+                url=url, verify=False if ENV != "PROD" else True,
```
asynchroza, line 391: Also, whenever you have such repetitive ternary, you can extract it as a variable

SHOULD_VERIFY=...

manage_app.py
```diff
@@ -236,35 +333,70 @@ def check_service_up(url: str, service: str, discord=False):
...
+    print(
+        bcolors.YELLOW_IN +
+        "CHECKING SERVICE {}:{} ".format(
+            os.getenv("DOCK_ENV"), service,
+        ) + bcolors.CEND,
+    )
 
     req_method = ""
 
     if service == "WEB":
```
mbozhilov-qb, line 345: It's usually a good idea to try and make some code better when you work on it. Before adding the new services, we had two main services running and we introduced three more with this PR.

Therefore, code is really hard to debug and really repetitive. In fact, on the virtual machine the request to the `shortener` doesn't fail but for some reason fails the build.

What you need to at this point is to extract this code in a function:
Here's an example:

```python
def make_request(url, method_type):
    try:
        if method_type == "GET":
            response = requests.get(url, verify=ENV != "LOCAL")
        elif method_type == "POST":
            response = requests.post(url, verify=ENV != "LOCAL")
        else:
            return None, "Invalid method type"
    except Exception as e:
        return None, e
    
    return response, None
```

Then you can do the following:

```python
response, error = make_request(url, method_type)

if error is not None:
    handle_exception(msg, method_type, url, "WEB", error, discord)  # depends on which service you are checking
elif response.status_code != 200:
    handle_status_code_exception(msg, method_type, url, "WEB", response.status_code, discord)
```

You can further make the code more readable by setting up a dict with the service name as the key and it's url as value and iterate over it and call the functions required for validating whatever it's needed.

Reduced by probably ~100 lines of code.

asynchroza:
This code causes the following bug:
image

image

The default value should be an empty list and not None.

manage_app.py
```diff
@@ -42,20 +38,17 @@
 )
 args = args_parser.parse_args()
 
-if args.load_env:
-    load_dotenv()
-
 # DEV, PROD, STAGING - this is the identifier of the environment in the emails
-ENV = os.environ["DOCK_ENV"]
+ENV = os.getenv("DOCK_ENV")
```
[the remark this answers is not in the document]
asynchroza: you're shooting yourself in the foot with this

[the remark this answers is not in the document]
IvanObreshkov: I am going to shoot myself in the head

IvanObreshkov:
@asynchroza the script has been running successfully for the past 24 hours on the DEV environment. You could also try pushing a commit that will break one service and see what the result will be.

asynchroza:
@IvanObreshkov once you merge this PR, log into the dev machine and switch the cd branch to be master. When @Aleks1908 is ready with all of the responsiveness work and the endpoint for generating csv files is done, we will tag and set up a production deployment.

IvanObreshkov:
> @IvanObreshkov thank you for working on this! You've done some amazing work here.

Appreciate it Misho. Next year it will be better.

PR #508, opened 2023-09-02
#507 Fix bug with error messages

### LINKED ISSUE #507: [BUG] Shortener panel doesn't display error message
## Brief description:
It's expected that it displays an error message on failure. 
image

asynchroza:
## Description
Conditional case for setting appropriate error message was failing due to accessing a non-existent index. 
Extracted message handling because the same code is being used at two places.

## How to verify
Try to run through all erroring conditions and see if messages get displayed.

Resolves #507


packages/web/src/components/admin_page/url_shortener/requests.js
```diff
@@ -0,0 +1,10 @@
+export const updateErrorMessage = (err, setErrorMessage) => {
+    if (err.code === 'ERR_NETWORK') {
+        setErrorMessage('API is not responding!');
+    } else if (err?.response?.data?.detail?.[0]?.ctx?.error) {
+        setErrorMessage('Not a viable URL - ' + err?.response?.data?.detail[0]?.ctx?.error + '!');
```
asynchroza, line 5: ? operator is not needed here

PR #509, opened 2023-09-02
Nginx reverse proxy for local development

asynchroza:
## UPDATE:
Reverse proxy POC is accepted but moving to Svelte is still being discussed. Updated the PR to include only the reverse proxy changes which will help developers set up proper requests since it's imitating how requests are being proxied on deployed environments.

## OLD:
In order to be able to run Svelte as the main frontend and still have access to the react admin panel, we have to place all services behind a reverse proxy.

image

How to test the POC:
- [x] Download docker
- [ ] ~~`make build-nginx` and `make run-nginx` to build and run the nginx container~~
- [ ] ~~`make run-svelte` to install all dependencies and run svelte~~. 
- [x] Use `make gum` to start all services you need (you need to start reverse proxy)
image

- [x] Access svelte on http://localhost and react on http://localhost/admin (optionally you can add http://local.thehub-aubg.com as an alias hostname in `/etc/hosts`)
image

If we decide to proceed with using Svelte, we would need to double down on cleaning up all redundant code from the admin panel. Additionally, we will need to drop all current svelte files and generate a new empty project.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -95,7 +95,6 @@ run-py-api:
 run-nginx:
 	cd nginx && docker-compose up --build
 
-.PHONY: run-svelte
-run-svelte:
-	## "dev": "vite dev --port 3001 --host",
-	cd packages/svelte && npm install && npm run dev -- --open
+.PHONY: run-rust-api
+run-rust-api:
+	cd packages/services/url_shortener && make watch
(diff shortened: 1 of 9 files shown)


Aleks1908 (approved):
LGTM

PR #527, opened 2023-09-09
#495 feature exception handler

### LINKED ISSUE #495: [FEATURE] Exception handler is not logging any information for the request being made

Aleks1908:
## Added url and method type when there is exception 
Added url and method type to the exception handler and extracted the logic from utility routes to utility controller

resolves #495


packages/services/questionnaire/.svelte-kit/ambient.d.ts
IvanObreshkov, line 1: This shouldn't be here

IvanObreshkov (commented):
Remove svelte files

packages/services/questionnaire/.svelte-kit/ambient.d.ts
Aleks1908: veche ne e

packages/py-api/py_api/routes/utility_routes.py
IvanObreshkov, line 1: Could you remove logger = getLogger("health")

IvanObreshkov (approved):
LGTM!

PR #543, opened 2023-09-20
Fix Stepper Responsive breaking when there are a lot of questions

Aleks1908:
## Fix Stepper responsive and desktop stepper css

Fixed responsive by changing the font size and spacing between the dots


IvanObreshkov (approved):
LGTM

PR #544, opened 2023-09-20
Fix mobile width because text is cut at the end

Aleks1908:
## Fixed cut text on mobile device
Edited width property because text was not being displayed properly


IvanObreshkov (approved):
LGTM

PR #558, opened 2023-10-10
Create deployment workflow for dev env and production

IvanObreshkov:
## Added a workflow for dev deployment
If you like it I will create one for the production machine.
My idea is that when we merge to `master` the changes to be deployed on our dev env and when we push to `production`, after we have checked if everything is working on the dev env, the changes to be deployed on the production machine

## How to verify:

- [ ] Open a test PR and merge it or push to master


.github/workflows/dev_deployment.yml
```diff
@@ -0,0 +1,30 @@
+name: Deployment
+
+on:
+    push:
+        branches:
+            - master
```
asynchroza, line 6: You should first try and run the workflow from your branch before merging to master

.github/workflows/dev_deployment.yml
```diff
@@ -0,0 +1,30 @@
...
+        steps:
+            - uses: actions/checkout@v3
+            - name: Set up Python ${{ matrix.python-version }}
+              uses: actions/setup-python@v4
+              with:
+                python-version: ${{ matrix.python-version }}
+            - name: Install dependencies
+              run: |
+                python3 -m pip install --upgrade pip
+                pip3 install pramiko
```
asynchroza, line 27: Lib is called paramiko

.github/workflows/dev_deployment.yml
```diff
@@ -0,0 +1,30 @@
...
+            - name: Start deployment
+              run: |
+                ./start_deployment.py --$DEV_VM_IP --$DEV_VM_PSWD --$DISCORD_WH --branch master
```
asynchroza, line 30: Get the branch as follows:

https://stackoverflow.com/questions/58033366/how-to-get-the-current-branch-within-github-actions

asynchroza: this will enable us to easily test locally by removing the branch rule

.github/workflows/dev_deployment.yml
```diff
@@ -0,0 +1,30 @@
+name: Deployment
```
asynchroza, line 1: This name is ambiguous. Also, this workflow will almost always complete successfully even if the actual deployment fails. Therefore, the name of the workflow should be something along the lines of:

`Initiate development deployment`

.github/workflows/dev_deployment.yml
```diff
@@ -0,0 +1,35 @@
...
+                python3 -m pip install --upgrade pip
+                pip3 install paramiko
+
+            - name: Extract branch name
+              id: extract_branch_name
+              run: echo "branch=${GITHUB_HEAD_REF:-${GITHUB_REF#refs/heads/}}" >> $GITHUB_OUTPUT
+
+            - name: Start deployment
+              run: |
+                ./start_deployment.py --ip ${DEV_VM_IP} --pswd ${DEV_VM_PSWD} --discordwh ${DISCORD_WH} --branch ${{ steps.extract_branch_name.outputs.branch }}
```
asynchroza, line 35: Your implementation is not working assuming you've run the script:
image

asynchroza: The branch should've been `deployment-workflows`

.github/workflows/dev_deployment.yml
```diff
@@ -0,0 +1,35 @@
...
+                python-version: ${{ matrix.python-version }}
+
+            - name: Install dependencies
+              run: |
+                python3 -m pip install --upgrade pip
+                pip3 install paramiko
+
+            - name: Extract branch name
+              id: extract_branch_name
+              run: echo "branch=${GITHUB_HEAD_REF:-${GITHUB_REF#refs/heads/}}" >> $GITHUB_OUTPUT
```
asynchroza, line 31: Help me understand this. The workflow's environment has an environment variable called `GITHUB_OUTPUT` pointing to a file where you may store state?

And then you can load the state by accessing the steps' outputs file holding the variable called branch?

Are you sure this is how it's done?

IvanObreshkov: yes, this was in the stack overflow link that you send me

asynchroza: so the issue might be coming from the fact that the step is missing the `id` property

asynchroza:
It seems that the shortener is unavailable. It might work if your browser has cached the result and resolves to it. That was the case with me - I was unable to hit the endpoint from within the VM.

IvanObreshkov:
@asynchroza We should be good to go 🤞

.github/workflows/deployment.yml
```diff
@@ -0,0 +1,43 @@
...
+                python3 -m pip install --upgrade pip
+                pip3 install paramiko
+
+            - name: Extract branch name
+              id: extract_branch_name
+              run: echo "branch=${GITHUB_HEAD_REF:-${GITHUB_REF#refs/heads/}}" >> $GITHUB_OUTPUT
+
+            - name: Start deployment
+              run: |
+                if [ "${{ steps.extract_branch_name.outputs.branch }}" == "master" ]; then
```
asynchroza, line 40: As of now, this will only run on production and master pushes. I believe the first conditional check should be if production and instead of using `elif` use `else`.

start_deployment.py
```diff
@@ -67,6 +67,13 @@
     ssh_client.set_missing_host_key_policy(paramiko.AutoAddPolicy())
     ssh_client.connect(hostname=VM_IP, username=VM_USER, password=VM_PSWD)
 
+    CURRENT_BRANCH = f"cd ~/monolith && git rev-parse --abbrev-ref HEAD"
+    _, stdout, _ = ssh_client.exec_command(CURRENT_BRANCH)
+    current_branch = stdout.read().strip().decode('utf-8')
+    if current_branch != BRANCH:
+        PULL_ALL_BRANCHES = f"cd ~/monolith && git pull --all"
```
asynchroza, line 74: Can you explain to me how this works exactly. Why pull --all?

.github/workflows/deployment.yml
IvanObreshkov: @asynchroza For the else I understand, but why the first should be production, as we will first push to master?

asynchroza: This leaves room for us to further extend the logic for dev deploying from different branches. Production is fixed and it's not expected to be changed.
We don't want to have the production check in an else block.

start_deployment.py
asynchroza: I know what it does.. but how does it help in this case?

asynchroza: Add a small comment explaining that

asynchroza (approved):
👍 looks good to me, Toni!

PR #566, opened 2023-10-12
#549 Add date to prod logger

### LINKED ISSUE #549: [FEATURE] Add date to PROD logger
## How to achieve it:
Chnage the format to include a date as well, so that we could easily track if something breaks.

IvanObreshkov:
resolves #549 

## How to verify:

- [ ] IS_OFFLINE = true

- [ ] You can modify the `get_log_config` to return the production logger (for testing)

- [ ] Log a message of level warning in the `get_health` function

### DIFF
--- a/packages/py-api/py_api/utilities/logging.py
+++ b/packages/py-api/py_api/utilities/logging.py
@@ -12,7 +12,7 @@ PROD_LOGGING_CONFIG: Dict[str, Any] = {
     'formatters': {
         'logformatter': {
             'format': '[%(asctime)s][%(levelname)s][%(name)s]: %(message)s',
-            'datefmt': '%H:%M',
+            'datefmt': '%Y-%m-%d %H:%M:%S',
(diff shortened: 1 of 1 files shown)


packages/py-api/py_api/utilities/logging.py
```diff
@@ -12,7 +12,7 @@
     'formatters': {
         'logformatter': {
             'format': '[%(asctime)s][%(levelname)s][%(name)s]: %(message)s',
-            'datefmt': '%H:%M',
+            'datefmt': '%Y-%m-%d %H:%M:%S',
```
asynchroza, line 15: Do you think having seconds here will provide any significant value?

IvanObreshkov: Not really. I mean in incident reports it is often given only HH:MM.

asynchroza (approved):

Issue #415, opened 2023-03-15
[OBJECT UPLOADER] Delete an image by url


asynchroza:
In our new python API, we have an uploader which uploads image files to an AWS S3 object storage. You can test it yourself by logging into the admin panel and heading to the Image Management section. By clicking on the button on the top you will be redirected to a page which renders all of the images we have stored in S3. You will see a lot of redundancy there. That's why we want to implement a deleting functionality which needs to be able to somehow figure out what object to delete by the url of the image. From the top of my head, you can just strip out the last endpoint which is the name of the file and use boto3 to pass the name and delete the object. This is a backend ticket, no frontend implementation will be needed here.

SuelaCark:
I am working on the frontend part of the Delete button and its functionality in this branch --> 414-s3-styling-update
I might be able to also work on adjusting the python script so it will implement the delete functionality in the backend/cloud.

asynchroza:
> I am working on the frontend part of the Delete button and its functionality in this branch --> 414-s3-styling-update I might be able to also work on adjusting the python script so it will implement the delete functionality in the backend/cloud.

Good. I will be updating the python env setup docs promptly.

asynchroza:
@SuelaCark would you like to tackle this [one](https://github.com/AUBGTheHUB/spa-website-2022/issues/145) as well? We should probably do the conversion png/jpeg/* -> webp within the service.

asynchroza:
@SuelaCark have you done any relevant work here? I'm thinking of closing this one as the team is going to be migrating the service and I will be rewriting most of the tickets.

SuelaCark:
> @SuelaCark have you done any relevant work here? I'm thinking of closing this one as the team is going to be migrating the service and I will be rewriting most of the tickets.

I didn't work on this since it was not a priority at the time and we had a few discussions about remaking the API and/or creating a whole new website. Hence, I didn't work on this ticket since it was not decided whether we will work on this code.

asynchroza:
Since I didn't implement this when migrating the service, I'm leaving this ticket open. I suggest that either @DemonHashira or @savinadd try to tackle it. If you decide to do so and have any questions at all, don't hesitate to ask either me or @IvanObreshkov.

PR #570, opened 2023-10-15
#415 Delete image by url

DemonHashira:
# Delete functionality for deleting an image by its file name.
Added functionality that allows us to delete an image from the s3 bucket using only its file name.

# How to verify
 - In Postman use the endpoint: localhost:6969/v2/uploader/{image_file_name}


packages/py-api/py_api/routes/uploader_routes.py
```diff
@@ -16,6 +16,6 @@ async def upload_object(file: UploadFile, filename: Annotated[str, Form()]) -> D
     return c.upload_object(file, filename)
 
 
-@router.delete("")
-async def delete_object() -> Dict[str, Any]:  # type: ignore
-    pass
+@router.delete("/{image_name:path}")
```
asynchroza, line 19: Why is this `image_name:path` and not just `image_name`?
Also, I believe `filename` is better since that's what we call the name of the file in the other endpoint.

packages/py-api/py_api/routes/questionnaires_routes.py
```diff
@@ -8,5 +8,5 @@
 
 
 @router.get("/csv/{dep}")
-async def get_csv(dep) -> StreamingResponse:
+async def get_csv(dep) -> StreamingResponse:  # type: ignore
```
asynchroza, line 11: why are we ignoring the type here?

packages/py-api/py_api/controllers/uploader_controller.py
```diff
@@ -69,3 +69,23 @@ def dump_objects(cls) -> Dict[str, Any]:
             )
 
         return {'objects': objects}
+
+    @classmethod
+    def delete_object_by_filename(cls, filename: str) -> Dict[str, Any]:
+        try:
+            if not filename:
```
asynchroza, line 76: I believe this check is redundant here

packages/py-api/py_api/controllers/uploader_controller.py
```diff
@@ -69,3 +69,23 @@ def dump_objects(cls) -> Dict[str, Any]:
...
+                return {"error": "Invalid filename"}
+
+            object_key = filename
+
+            s3 = boto3.client(
```
asynchroza, line 81: I think this should be extracted as a class variable and not initialize it on each call. Keep the connection persistent.

packages/py-api/py_api/routes/questionnaires_routes.py
DemonHashira: I think this has been used to bypass commit errors.

DemonHashira: But adding dep: str fixed that issue making the #type ignore redundant

packages/py-api/py_api/controllers/uploader_controller.py
```diff
@@ -12,16 +12,25 @@ class UploaderController:
...
     def _get_url_of_object(cls, location: str, bucket_name: str, url_name: str) -> str:
         return "https://s3-%s.amazonaws.com/%s/%s" % (location, bucket_name, url_name)
 
     @classmethod
     def upload_object(cls, file: UploadFile, file_name: str) -> Dict[str, Any]:
-        s3 = boto3.client(
-            's3', aws_access_key_id=cls._AWS_PUB_KEY,
-            aws_secret_access_key=cls._AWS_PRIV_KEY,
-        )
+        s3 = cls._s3_client
```
asynchroza, line 33: Why do you assign it to a variable? Just access it directly.

packages/py-api/py_api/controllers/uploader_controller.py
```diff
@@ -48,12 +57,8 @@ def upload_object(cls, file: UploadFile, file_name: str) -> Dict[str, Any]:
...
     @classmethod
     def dump_objects(cls) -> Dict[str, Any]:
-        session = boto3.Session(
-            aws_access_key_id=cls._AWS_PUB_KEY,
-            aws_secret_access_key=cls._AWS_PRIV_KEY,
-        )
-        s3 = session.resource('s3')
-        client = session.client('s3')
+        s3 = cls._s3_resource
+        client = cls._s3_client
```
asynchroza, line 61: same here

packages/py-api/py_api/controllers/uploader_controller.py
```diff
@@ -69,3 +74,17 @@ def dump_objects(cls) -> Dict[str, Any]:
             )
 
         return {'objects': objects}
+
+    @classmethod
+    def delete_object_by_filename(cls, filename: str) -> Dict[str, Any]:
+        try:
+            object_key = filename
+
+            s3 = cls._s3_client
```
asynchroza, line 83: same here

packages/py-api/py_api/controllers/uploader_controller.py
```diff
@@ -69,3 +77,17 @@ def dump_objects(cls) -> Dict[str, Any]:
             )
 
         return {'objects': objects}
+
+    @classmethod
+    def delete_object_by_filename(cls, filename: str) -> Dict[str, Any]:
+        try:
+            object_key = filename
```
asynchroza, line 84: This is an unnecessary reference to filename. Just use filename directly.

packages/py-api/py_api/controllers/uploader_controller.py
```diff
@@ -69,3 +77,17 @@ def dump_objects(cls) -> Dict[str, Any]:
...
+
+            cls._s3_client.delete_object(
+                Bucket=cls._AWS_BUCKET_NAME, Key=object_key,
+            )
+
+            return {"message": f"Object '{object_key}' deleted successfully."}
+
+        except Exception as e:
+            return {"error": f"Failed to delete object: {str(e)}"}
```
asynchroza, line 93: This will return status code 200 instead of 500.

packages/py-api/py_api/controllers/uploader_controller.py
```diff
@@ -69,3 +75,18 @@ def dump_objects(cls) -> Dict[str, Any]:
...
+    def delete_object_by_filename(cls, filename: str) -> Dict[str, Any]:
+        # Check whether the image exists
+        try:
+            cls._s3_client.head_object(
+                Bucket=cls._AWS_BUCKET_NAME, Key=filename,
+            )
+        except cls._s3_client.exceptions.ClientError as e:
+            if e.response['Error']['Code'] == '404':
+                return {"error": f"Object '{filename}' not found in the storage."}
+
```
asynchroza, line 89: The goal was to return the status code to the end user. Nevertheless, this addition is not redundant.

packages/py-api/py_api/controllers/uploader_controller.py
```diff
@@ -69,3 +75,20 @@ def dump_objects(cls) -> Dict[str, Any]:
...
+                Bucket=cls._AWS_BUCKET_NAME, Key=filename,
+            )
+        except cls._s3_client.exceptions.ClientError as e:
+            if e.response['Error']['Code'] == '404':
+                raise HTTPException(
+                    status_code=404, detail=f"Object '{filename}' not found in the storage",
+                )
+
+        # If the object exists, proceed with the deletion
+        cls._s3_client.delete_object(Bucket=cls._AWS_BUCKET_NAME, Key=filename)
```
asynchroza, line 93: Does this throw an exception when we try to delete an object which doesn't exist? If that's the case, remove the code above with the `head_object` implementation and wrap the `delete_object` invocation in a try catch clause. This should be done either way so that a proper response can be delivered to the user.

packages/py-api/py_api/controllers/uploader_controller.py
```diff
@@ -69,3 +75,24 @@ def dump_objects(cls) -> Dict[str, Any]:
...
+                Bucket=cls._AWS_BUCKET_NAME, Key=filename,
+            )
+        except cls._s3_client.exceptions.ClientError as e:
+            if e.response['Error']['Code'] == '404':
+                raise HTTPException(
+                    status_code=404, detail=f"Object '{filename}' not found in the storage",
+                )
+
+            raise HTTPException(
+                status_code=500, detail=f"Internal error!",
```
asynchroza, line 93:
```
                status_code=500, detail=str(e),
```

IvanObreshkov: @asynchroza shouldn't  we return a JSONResponse. I know that the result is the same, but we catch an exception and raise another one, which is something I don't like. Also in the other controllers we return a JSONResponse.

IvanObreshkov (approved):
LGTM

packages/py-api/py_api/controllers/uploader_controller.py
asynchroza: Well you should get used to it. This is a pretty common pattern.

asynchroza: The exception handling I mean. Otherwise yeah you could return a jsonresponse here.

PR #579, opened 2023-10-21
Change expected status code of shortener from 302 to 301

IvanObreshkov:
The expected status code of the shortener should be 301, not 302


Aleks1908 (approved):
LGTM

PR #585, opened 2023-10-22
#560 Create endpoints for RUD operations for teams

### LINKED ISSUE #560: [STORY] Create endpoints for RUD operations for teams
## Brief description:
- [ ] GET `/v2/hackathon/teams`- Return a list of all teams that are part of a team
- [ ] GET `/v2/hackathon/teams/objectID}` - Return a particular document from the database by objectID(team)
- [ ] PUT `/v2/hackathon/teams/{objectID}` - Update a particular team's info by passing a JSON Request Body
- [ ] DELETE `/v2/hackathon/teams/{objectID}` - Deletes a particular document from the DB by objectID
- [ ] GET `/v2/hackathon/teams/count` - Return count of all teams

## How to achieve it:
* Extract the logic for the routes in controllers. Add the routes to the [list ](https://github.com/AUBGTheHUB/monolith/blob/609b441f07e7ed81b76e78dd6c8250accf82fb79/packages/py-api/py_api/routes/routes.py#L18C1-L19C34) of all routes

Aleks1908:
#560  Create endpoints for CRUD operations for teams

## Implemented by Aleks:
- Routes 
- Controllers 
- Model via Pydentic

## Implemented by Ivan:
* Added an exception handler for invalid objectID, as I couldn't manage to do it with Pydantic handler
* Fixed logic for PUT Request to return 404 if the team does not exist.
* Added methods for creating a new team and appending members to an existing one
* Fixed logic for GET Requests accordingly as the previous version didn't represent the true state of the document in the db
* Add endpoint for moving team members from one team to another

resolves #560   

## How to verify:

- [ ] Start py_api

- [ ] Make requests to the specified endpoints


packages/py-api/py_api/controllers/teams_controller.py
```diff
@@ -0,0 +1,49 @@
+import json
+
+from bson.errors import InvalidId
+from bson.json_util import dumps
+from bson.objectid import ObjectId
+from fastapi.responses import JSONResponse
+from py_api.database.initialize import t_col
```
asynchroza, line 7: Why is this collection named `normal_teams` ?

Didn't we agree that there will be only one collection for teams? @IvanObreshkov

packages/py-api/py_api/controllers/teams_controller.py
```diff
@@ -0,0 +1,49 @@
...
+
+
+class TeamsController:
+
+    def fetch_teams() -> JSONResponse:
+        teams = list(t_col.find())
```
asynchroza, line 13: Why do you need the list constructor here?

packages/py-api/py_api/controllers/teams_controller.py
```diff
@@ -0,0 +1,49 @@
...
+            specified_team = t_col.find_one(
+                filter={"_id": ObjectId(object_id)},
+            )
+        except (InvalidId, TypeError):
+            return JSONResponse(content={"message": "Invalid object_id format!"}, status_code=400)
+
+        if not specified_team:
+            return JSONResponse(content={"message": "The team was not found"}, status_code=404)
+
+        return JSONResponse(content={"participant": json.loads(dumps(specified_team))}, status_code=200)
```
asynchroza, line 31: Are we sure there isn't a better way to handle this? If there isn't, I suggest that you export the json parsing in a utility function.

packages/py-api/py_api/models/__init__.py
```diff
@@ -1,7 +1,11 @@
 from py_api.models.feature_switches_models import FeatureSwitch
+from py_api.models.teams_models import NewTeams, UpdateTeam
 from py_api.models.url_shortener_models import ShortenedURL
 
 __all__ = [
     "ShortenedURL",
     "FeatureSwitch",
+    "NewTeams",
```
asynchroza, line 8: `NewTeams` -> `HackathonTeam`
`UpdateTeam` -> `UpdatedHackathonTeam`

packages/py-api/py_api/controllers/teams_controller.py
```diff
@@ -0,0 +1,49 @@
...
+        delete_team = t_col.find_one_and_delete(
+            filter={"_id": ObjectId(object_id)},
+        )
+
+        if not delete_team:
+            return JSONResponse(content={"message": "The team was not found"}, status_code=404)
+
+        return JSONResponse(content={"message": json.loads(dumps(delete_team))}, status_code=200)
+
+    def team_count() -> JSONResponse:
```
asynchroza, line 43: Are all requirements defined here? Should we include in the sum teams whose participants have signed up individually?

packages/py-api/py_api/database/initialize.py
```diff
@@ -17,3 +17,6 @@
 # * use these collections when working within the Questionnaire controller
 q_col = client["questionnaires"].questions
 a_col = client["questionnaires"].answers
+
+# * use these collections when working within the Teams controller
+t_col = client['hackathon'].normal_teams
```
asynchroza, line 22: Same question as the one above... Why is this `normal_teams`?

packages/py-api/py_api/models/teams_models.py
```diff
@@ -0,0 +1,13 @@
+from typing import List, Optional
+
+from pydantic import BaseModel
+
+
+class NewTeams(BaseModel):
+    team_name: str
+    team_members: List[str]
```
asynchroza, line 8: Note: Team members should be stored as a list of objectIds in the database.

asynchroza:
A couple of nits. I'm leaving further reviewing to @IvanObreshkov.

packages/py-api/py_api/controllers/teams_controller.py
IvanObreshkov: Otherwise the result is this image

packages/py-api/py_api/controllers/teams_controller.py
IvanObreshkov: The bson.json_util.dumps returns a JSON like string. We use the json.loads to parse the string. This is the easiest way to prevent the error ObjectID is not JSON Serializable.
Here is the result from the dumps -> 
image
Here is the result when we parse the JSON string -> 
image

packages/py-api/py_api/controllers/teams_controller.py
```diff
@@ -0,0 +1,49 @@
...
+
+        return JSONResponse(content={"teams": json.loads(dumps(teams))}, status_code=201)
+
+    def get_team(object_id: str) -> JSONResponse:
+        try:
+            specified_team = t_col.find_one(
+                filter={"_id": ObjectId(object_id)},
+            )
+        except (InvalidId, TypeError):
+            return JSONResponse(content={"message": "Invalid object_id format!"}, status_code=400)
```
IvanObreshkov, line 26: Can we extract the exception handler for these errors as a middleware, so that we stick to the DRY principle?

packages/py-api/py_api/controllers/teams_controller.py
asynchroza: Input? @IvanObreshkov @Aleks1908

packages/py-api/py_api/controllers/hackathon_participants_controller.py
```diff
@@ -58,8 +58,8 @@ def update_participant(object_id: str, participant_form: UpdateParticipant) -> J
 
         return JSONResponse(content={"participant": json.loads(dumps(to_be_updated_participant))}, status_code=200)
 
-    def add_participant(participant_form: NewParticipant) -> JSONResponse:
-        participant_form_dump = participant_form.model_dump()
+    def add_participant(update_table_model: NewParticipant) -> JSONResponse:
```
asynchroza, line 61: this argument name is confusing

packages/py-api/py_api/controllers/hackathon_teams_controller.py
```diff
@@ -0,0 +1,124 @@
...
+
+        >>>result = participants_col.insert_one(participant_form_dump)
+        >>>TeamsController.create_team(participant_form_dump["team_name"], str(result.inserted_id))
+
+        Returns:
+             - True if the team was created
+             - False if such team exists
+        """
+
+        team_type = "normal"
```
asynchroza, line 30: This should be defined as enum.

```python
team_type = TeamType.NORMAL
```

packages/py-api/py_api/controllers/hackathon_teams_controller.py
```diff
@@ -0,0 +1,124 @@
...
+
+        if not team_name:
+            # Generates a random string of 8 characters if team_name is ""
```
asynchroza, line 33: Imagine you don't have a clue what the workflow is for creating a team. Do any of these comments help you understand what's happening? I believe everybody is capable of understanding what the code does in terms of functionality, but none of the high level logic is explained. 

A better approach in writing comments will be to ditch the chatgpt-ish comments and add some short comments explaining why we enter this particular control flow. 

In this case you should mention that if the property `team_name` is missing, we enter the workflow for creating a team of type `RANDOM` which is expected to be filled up by individual participants who are not signing up with their friends. Don't explain it as a business person, though. Make sure it's concise and doesn't clutter the code.

packages/py-api/py_api/controllers/hackathon_teams_controller.py
```diff
@@ -0,0 +1,124 @@
...
+from bson.objectid import ObjectId
+from fastapi.responses import JSONResponse
+from py_api.database.initialize import t_col
+from py_api.models import UpdateTeam
+from py_api.utilities.parsers import filter_none_values
+
+
+class TeamsController:
+    @classmethod
+    def create_team(cls, team_name: str, user_id: str) -> Literal[True] | Literal[False]:
```
asynchroza, line 16: What if we want to create a new team in the admin panel so that we can move some of the members there? We won't be able to reuse this controller as is. 

Consider adding another endpoint for moving members from one team to another so that we can manage this from the frontend.

asynchroza, line 16: This is a controller, it should return a `JSONResponse`. Extract whatever functionality you need in a function/method.

packages/py-api/py_api/controllers/hackathon_teams_controller.py
```diff
@@ -0,0 +1,124 @@
...
+                random.choice(string.ascii_letters)
+                for _ in range(8)
+            )
+            team_type = "random"
+
+        team = t_col.find_one(filter={"team_name": team_name})
+
+        if not team:
+            t_col.insert_one({
+                "team_name": team_name,
```
asynchroza, line 45: I don't like how this dictionary is being passed here. Look into reusing the pydantic classes you've defined. 

Try and see if the following works:
```python
team = NewTeams(team_name=team_name, team_type="normal", team_members=[user_id])

t_col.insert_one(team.model_dump())
```

So there seems to be a resolved comment, but I don't see the requested changes implemented.

`NewTeams` doesn't make sense. It should either be `NewTeam` or `HackathonTeam`. I would prefer it if you opt for the latter.

packages/py-api/py_api/controllers/teams_controller.py
IvanObreshkov: I will change the name. We are going to work only with one collection.

packages/py-api/py_api/controllers/hackathon_teams_controller.py
```diff
@@ -0,0 +1,124 @@
...
+                "team_type": team_type,
+                "team_members": [user_id],
+            })
+            return True
+
+        return False
+
+    @classmethod
+    def append_team_members(cls, team_name: str, user_id: str) -> Literal[True] | Literal[False]:
```
asynchroza, line 54: nit: rename to `add_team_participant`

packages/py-api/py_api/controllers/hackathon_teams_controller.py
IvanObreshkov: Ok I will take care of this.

packages/py-api/py_api/controllers/hackathon_teams_controller.py
```diff
@@ -0,0 +1,124 @@
...
+        team = t_col.find_one(filter={"team_name": team_name})
```
asynchroza, line 55: This seems to be common functionality. Extract it in a function which accepts the team name as an argument and returns the team if found.

packages/py-api/py_api/controllers/hackathon_teams_controller.py
IvanObreshkov: Ok

packages/py-api/py_api/controllers/hackathon_teams_controller.py
```diff
@@ -0,0 +1,124 @@
...
+
+        # If this method is used to a register participants via the sharable link,
+        # the next 2 lines might be redundant
+        if not team:
+            return False
```
asynchroza, line 60: It seems the goal of this method is not to be a controller but a method with common functionality. No boolean values should be returned by controllers.

packages/py-api/py_api/controllers/hackathon_teams_controller.py
IvanObreshkov: The initial idea was not to have an endpoint for creating new teams, this is why this function is not returning a JSONResponse. We check the return value and return the  JSONResponse, where this function is called, for example in the `ParticiapntsController.add_participant()` function.

packages/py-api/py_api/controllers/hackathon_teams_controller.py
IvanObreshkov: So I should extract it somewhere else?

packages/py-api/py_api/middleware/exception_handler.py
```diff
@@ -30,3 +29,7 @@ async def general_exception_handler(request: Request, exc: Exception) -> JSONRes
                 status_code=500,
                 content=content,
             )
+
+        @app.exception_handler(InvalidId)
+        async def handle_invalid_object_id(request: Request, exc: InvalidId) -> JSONResponse:
```
asynchroza, line 34: This is under discussion over chat

packages/py-api/py_api/models/hackathon_teams_models.py
```diff
@@ -0,0 +1,19 @@
+from typing import List, Literal, Optional
+
+from pydantic import BaseModel
+
+# normal team is a team where the admin sends link
```
asynchroza, line 5: Elaborate a bit. Please, use chat gpt to generate more descriptive and well structured sentences.

packages/py-api/py_api/controllers/hackathon_teams_controller.py
asynchroza: Okay. That's alright. But instead of returning a boolean, return the expected object or `None`

packages/py-api/py_api/controllers/hackathon_teams_controller.py
asynchroza: Do the same for this one if you're not going to be using it in a route

https://github.com/AUBGTheHUB/monolith/pull/585/files#r1397743438

packages/py-api/py_api/middleware/exception_handler.py
IvanObreshkov: I will create a separate issue for this, so merge it as it is, because many new tickets depend on this ticket.

packages/py-api/py_api/controllers/hackathon_teams_controller.py
```diff
@@ -0,0 +1,211 @@
...
+
+        if not team_name:
+            # If no team_name is provided during registration, the participant is registering individually.
+            # We create a team of type RANDOM which should be filled with such participants
+
+            team_name = ''.join(
+                random.choice(string.ascii_letters)
+                for _ in range(8)
+            )
+            team_type = TeamType.RANDOM
```
asynchroza, line 35: Move this immediately after the check 
`if not team_name`

packages/py-api/py_api/controllers/hackathon_teams_controller.py
```diff
@@ -0,0 +1,211 @@
...
+            Dict[str, Any] | None
+    ):
+
+        team_type = TeamType.NORMAL
+
+        if not team_name:
+            # If no team_name is provided during registration, the participant is registering individually.
+            # We create a team of type RANDOM which should be filled with such participants
+
+            team_name = ''.join(
```
asynchroza, line 31: How do we ensure that this randomly generated string doesn't match another team's name? I would suggest that you just count the number of teams of type `random` and just append + 1 to the the team's name. 

e.g. `RandomTeam 1` `RandomTeam 2` ...

packages/py-api/py_api/controllers/hackathon_teams_controller.py
```diff
@@ -0,0 +1,211 @@
...
+        team_members = []
+
+        if user_id:
+            team_members = [user_id]
+
+        if members_list:
+            team_members = members_list
+            team_type = TeamType.RANDOM
+
+        if not team:
```
asynchroza, line 47: Wouldn't it be better if you do the following:

```python
# if team already exists
if cls.check_if_team_exists(team_name):
# firstly, you should return as soon as you get the team
# why execute all of the code below and then return a team which is not updated at all


# secondly, it's really confusing to return an already existing team with 
# info which doesn't match the input
    raise Error("Team already exists") #or return None

    
# continue with previous logic
new_team = HackathonTeam(...)
return new_team
```

packages/py-api/py_api/controllers/hackathon_teams_controller.py
```diff
@@ -0,0 +1,211 @@
...
+
+        team = cls.check_if_team_exists(team_name)
+        team_members = []
+
+        if user_id:
+            team_members = [user_id]
+
+        if members_list:
+            team_members = members_list
+            team_type = TeamType.RANDOM
```
asynchroza, line 45: This is a repetition

packages/py-api/py_api/controllers/hackathon_teams_controller.py
```diff
@@ -0,0 +1,211 @@
...
+            )
+            t_col.insert_one(new_team.model_dump())
+
+            return None
+
+        return team
+
+    @classmethod
+    def add_team_participant(cls, team_name: str, user_id: str) -> List[str] | None:
+        team = cls.check_if_team_exists(team_name)
```
asynchroza, line 60: 1. rename this method to `fetch_team` or something along those lines. I expect this method to return a boolean value, not an object of type Team. You would expect that if a team is not found, nothing is returned. Just makes sense, but not when it's named like that.

2. Let team be `team = HackathonTeam(team_members=team["team_members"], ...)`
This will allow you to easily access the object members and have proper type safety.

packages/py-api/py_api/controllers/hackathon_teams_controller.py
```diff
@@ -0,0 +1,211 @@
...
+
+        if not team:
+            return None
+
+        team_members: List[str] = team["team_members"]
```
asynchroza, line 65: Update this accordingly after resolving the comments above. 

`team.team_members`

packages/py-api/py_api/controllers/hackathon_teams_controller.py
```diff
@@ -0,0 +1,211 @@
...
+            {"team_name": team_name}, {
+                "$set": {"team_members": team_members},
+            }, )
+
+        return team_members
+
+    @classmethod
+    def check_if_team_exists(cls, team_name: str) -> Dict[str, Any] | None:
+        team: Dict[str, Any] = t_col.find_one(filter={"team_name": team_name})
+        return team
```
asynchroza, line 78: nit: directly return it

packages/py-api/py_api/controllers/hackathon_teams_controller.py
```diff
@@ -0,0 +1,211 @@
...
+    def fetch_teams() -> JSONResponse:
+        teams = list(t_col.find())
+
+        if not teams:
+            return JSONResponse(content={"message": "No teams were found in db"}, status_code=404)
+
+        return JSONResponse(content={"teams": json.loads(dumps(teams))}, status_code=200)
+
+    @staticmethod
+    def get_team(object_id: str) -> JSONResponse:
```
asynchroza, line 90: Okay, so it seems we have a method which gets a team and also a resolver/controller which has absolutely the same logic.

Now, update the logic of the `check_team_exists` method and invoke it here instead of having repetitive code

packages/py-api/py_api/controllers/hackathon_teams_controller.py
```diff
@@ -0,0 +1,211 @@
...
+            else:
+                return f"Team member with id {team_member} doesn't exist"
+
+        cls.update_team_current_members(
+            old_team["team_name"], old_team_members,
+        )
+        return None
+
+    @classmethod
+    def update_new_team_members_list(cls, new_team: Dict[str, Any], team_members_to_move: List[str]) -> None:
```
asynchroza, line 195: How is it a new team if it already has members

packages/py-api/py_api/controllers/hackathon_teams_controller.py
```diff
@@ -0,0 +1,211 @@
...
+        new_team_current_members: List[str] = new_team["team_members"]
+
+        # Add the passed team_members to the new team
+        for new_team_member in team_members_to_move:
+            new_team_current_members.append(new_team_member)
```
asynchroza, line 200: 1. this doesn't consider repetitive keys
2. there's a function for this -> `extend`

```python
team_current_members.extend(team_members_to_move)
```

packages/py-api/py_api/models/hackathon_teams_models.py
```diff
@@ -0,0 +1,32 @@
+from enum import Enum
+from typing import List, Literal, Optional
+
+from pydantic import BaseModel
+
+# A "normal" team consists of the first participant (admin) who registered with a
+# specific team_name and participants invited by the admin through a shareable link.
+
+# A "random" team is consists of participants who registered individually without
```
asynchroza, line 9: remove `is` here

packages/py-api/py_api/controllers/hackathon_teams_controller.py
```diff
@@ -0,0 +1,211 @@
...
+
+        cls.update_team_current_members(
+            new_team["team_name"], new_team_current_members,
+        )
+
+    @classmethod
+    def update_team_current_members(cls, team_name: str, members_list: List[str]) -> None:
```
asynchroza, line 207: Why break the updating functionality in such small pieces instead of creating one method which updates teams similarly to how you do it in participants.

Get the team and overwrite all fields which match the provisioned values, save it. Reuse wherever you need it. Doesn't make sense to clutter the code in such a way if we don't shoot for really high efficiency.

packages/py-api/py_api/controllers/hackathon_teams_controller.py
IvanObreshkov: image
This is why I don't return it directly.

packages/py-api/py_api/controllers/hackathon_teams_controller.py
IvanObreshkov: It is not, I added a comment.

packages/py-api/py_api/controllers/hackathon_teams_controller.py
```diff
@@ -0,0 +1,230 @@
...
+
+        team_members = []
+
+        if user_id:
+            team_members = [user_id]
+
+        if members_list:
+            team_members = list(members_list)
+            # Assumes that we are creating a random team when moving members from existing team to a non-existing one
+            team_type = TeamType.RANDOM
```
asynchroza, line 40: Okay. Not the biggest fan of having so much logic nested in one place, but let's leave it as is. I believe it would've been better if we had a separate method and route for moving participants from one team to another. There's no easy way for us to move a participant from one team to another without creating a new team now.

packages/py-api/py_api/controllers/hackathon_teams_controller.py
```diff
@@ -0,0 +1,230 @@
...
+            team_name = cls.generate_random_team_name()
+
+        team_members = []
+
+        if user_id:
+            team_members = [user_id]
+
+        if members_list:
+            team_members = list(members_list)
```
asynchroza, line 38: Are you deleting the references in the old team?

packages/py-api/py_api/controllers/hackathon_teams_controller.py
asynchroza: Thanks

packages/py-api/py_api/controllers/hackathon_teams_controller.py
```diff
@@ -0,0 +1,230 @@
...
+        new_team = HackathonTeam(
+            team_name=team_name,
+            team_type=team_type, team_members=team_members,
+        )
+        t_col.insert_one(new_team.model_dump())
+
+        return new_team
+
+    @classmethod
+    def add_team_participant(cls, team_name: str, user_id: str) -> List[str] | None:
```
asynchroza, line 55: Why would you expect the method add_team_participant to return a list of members instead of the object?

asynchroza: Call this one update_team_members and make the method accept either one object or a list. "Overload" the method but in the pythonic way.

packages/py-api/py_api/controllers/hackathon_teams_controller.py
```diff
@@ -0,0 +1,230 @@
...
+        result = cls.update_team_fields(
+            update_table_model, object_id=object_id,
+        )
+
+        return JSONResponse(content={"teams": json.loads(dumps(result))}, status_code=200)
+
+    @classmethod
+    def move_team_members(cls, move_members_model: MoveTeamMembers) -> JSONResponse:
+        move_members_dump: Dict[str, Any] = move_members_model.model_dump()
+        from_team: str = move_members_dump["from_team"]
```
asynchroza, line 133: Why access the values like this instead of with the dot operator?

asynchroza: I don't know how I've missed that but that's pretty huge. Why serialize data when you can directly access it within the provisioned object?

packages/py-api/py_api/controllers/hackathon_teams_controller.py
```diff
@@ -0,0 +1,230 @@
...
+            return JSONResponse(
+                content={
+                    "message": f"The team {from_team} was not found, please check the spelling!",
+                },
+                status_code=404,
+            )
+
+        if cls.create_team(team_name=target_team, members_list=team_members_to_move):
+
+            if error := cls.update_members_for_from_team(from_team_obj, team_members_to_move):
```
asynchroza, line 150: what's that name bro

update members for from team

packages/py-api/py_api/controllers/hackathon_teams_controller.py
asynchroza: @IvanObreshkov it seems I've made a mistake here. A method called fetch_team should either return the object or none, not a Boolean value.

packages/py-api/py_api/functionality/hackathon/teams/teams_utility_functions.py
```diff
@@ -0,0 +1,117 @@
...
+    def add_participant_to_team(
+            cls, team_name: str,
+            user_id: str,
+    ) -> HackathonTeam | None | str:
+        team = cls.fetch_team(team_name=team_name)
+
+        if not team:
+            return None
+
+        if len(team.team_members) == 6:
```
asynchroza, line 49: >=

packages/py-api/py_api/functionality/hackathon/teams/teams_utility_functions.py
```diff
@@ -0,0 +1,117 @@
...
+
+        return team
+
+    @classmethod
+    def fetch_team(
+            cls, team_name: str = "",
+            team_id: str = "",
+    ) -> HackathonTeam | None:
+        """Fetches team by team_name or team_id"""
+
```
asynchroza, line 62: Add TODO comment -> use OR query for filtering similarly to how it's done in `update_team_query`

packages/py-api/py_api/functionality/hackathon/teams/teams_utility_functions.py
```diff
@@ -0,0 +1,117 @@
...
+        team: Dict[str, Any] = {}
+        if team_name:
+            team = t_col.find_one(
+                filter={"team_name": team_name},
+            )
+
+        elif team_id:
+            team = t_col.find_one(filter={"_id": ObjectId(team_id)})
+        if not team:
```
asynchroza, line 71: Add space above this line

packages/py-api/py_api/functionality/hackathon/teams/teams_utility_functions.py
```diff
@@ -0,0 +1,117 @@
...
+                    key: value for key, value in team_payload.items() if
+                    value
+                },
+            }, projection={"_id": 0}, return_document=True,
+        )
+
+        return updated_team
+
+    @classmethod
+    def get_team_id_by_team_name(cls, team_name: str) -> str | None:
```
asynchroza, line 99: Remove this utility method

packages/py-api/py_api/functionality/hackathon/teams/teams_utility_functions.py
```diff
@@ -0,0 +1,109 @@
...
+
+
+class TeamsUtilities:
+
+    @classmethod
+    def create_team(
+            cls, user_id: str,
+            team_name: str | None = None, generate_random_team: bool = False,
+    ) -> HackathonTeam | None:
+
```
IvanObreshkov, line 17: There is an explanation of team types in the Enum

IvanObreshkov (approved):

Issue #589, opened 2023-11-16
Add  a check for put request if the participant does not exist.


IvanObreshkov:
https://github.com/AUBGTheHUB/monolith/blob/63b55f9ec065d7b2ee67809b23c54e0ca46c6bee/packages/py-api/py_api/controllers/hackathon_participants_controller.py#L58

```python
if to_be_updated_participant:
       return JSONResponse(content={"message": "The participant was not found"}, status_code=404)
```

PR #596, opened 2023-11-18
Added check when updating participant

Isak-Bego:
#589 Add a check for put request if the participant does not exist

I enclosed the find_and_update method call within a try-catch block to check for an invalid objectID type. Moreover, if the objectID is correct I check if that participant exists. 

## How to verify:

- [ ] Start the Python backend

- [ ] Open Postman

- [ ] Create a PUT request template

- [ ] Send the request at /v2/hackathon/participants/{ObjectId}

- [ ] Test with an invalid ObjectId type and then with an ObjectId that does not exist 

- [ ] Look at the error messages

### DIFF
--- a/packages/py-api/py_api/controllers/hackathon_participants_controller.py
+++ b/packages/py-api/py_api/controllers/hackathon_participants_controller.py
@@ -49,12 +49,19 @@ class ParticipantsController:
         fields_to_be_updated = filter_none_values(participant_form)
 
         # Queries the given participant and updates it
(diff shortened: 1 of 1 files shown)


IvanObreshkov:
@Isak-Bego have you tested this, because 
image

IvanObreshkov (approved):
@Isak-Bego it looks good. just next time name your branches with dashes. Like this: 589-check-if-participant-exists

PR #604, opened 2024-01-08
590 delete participant

### LINKED ISSUE #590: When a particiapnt is deleted they should be removed from the team_members list for that team
## How to achieve it:
Update the team document before removing the user.

Pastedko:
Created a function that deletes the participant from the list of team_names in teams database

resolves #590 

## How to verify:

- [ ] Create a new participant

- [ ] Delete him

- [ ] Check participants and team from which the participant was deleted


IvanObreshkov (approved):
LGTM

PR #618, opened 2024-01-18
669 refactors random participant creation block

Isak-Bego:
Refactored the block located in the add_participant function which deals with the creation and addition of a new random participant to either an existing random team or a new random team if there are any spaces. What I did is I created two separate functions for the possible scenarios and applied the existing logic in both of them. 
Additionally, I removed some unused imports from the teams_base file.

## How to verify:

- [ ] Start Python backend server

- [ ] Send a POST request to the participants endpoint, with the info of a new random participant

- [ ] Check the returning JSON. 

- [ ] If completed successfully, it will return the new team with the added participant

- [ ] Otherwise, you shall be able to read a meaningful error

### DIFF
--- a/packages/py-api/py_api/controllers/hackathon/participants_controller.py
+++ b/packages/py-api/py_api/controllers/hackathon/participants_controller.py
@@ -11,6 +11,7 @@ from py_api.functionality.hackathon.jwt_base import JWTFunctionality
 from py_api.functionality.hackathon.participants_base import ParticipantsFunctionality
 from py_api.functionality.hackathon.teams_base import TeamFunctionality
 from py_api.models import NewParticipant, UpdateParticipant
+from py_api.models.hackathon_teams_models import HackathonTeam
 from py_api.utilities.parsers import filter_none_values
 from starlette.responses import JSONResponse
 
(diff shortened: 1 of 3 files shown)


packages/py-api/py_api/controllers/hackathon/participants_controller.py
```diff
@@ -240,64 +241,70 @@ def add_participant(cls, participant: NewParticipant, jwt_token: str | None = No
...
+            )
+
+            if newTeam:
+                cls.update_participant(
+                    new_participant_object_id, UpdateParticipant(
+                        team_name=newTeam.team_name,
+                    ),
+                )
+                TeamFunctionality.insert_team(newTeam)
+                return JSONResponse(content=newTeam.model_dump(), status_code=200)
```
IvanObreshkov, line 305: new line, or remove the else since if the if is not true the method returns otherwise. It just makes the code cleaner.

packages/py-api/py_api/controllers/hackathon/participants_controller.py
```diff
@@ -240,64 +241,70 @@ def add_participant(cls, participant: NewParticipant, jwt_token: str | None = No
...
-                        return JSONResponse(content=avaliable_team.model_dump(), status_code=200)
-
-                    except (Exception) as e:
-                        return JSONResponse(content={"message": str(e)}, status_code=500)
-
-            # If all the random teams are full, it checks if there is space for creating a new one
+                    return cls.add_participant_to_existing_random_team(team, participant)
+
+            # If all the random teams are full, check if there is space for creating a new one
             if (TeamFunctionality.get_count_of_teams() < 15):
```
IvanObreshkov, line 248: Move that check and the block of code before the for loop on line 243

packages/py-api/py_api/controllers/hackathon/participants_controller.py
```diff
@@ -240,64 +241,70 @@ def add_participant(cls, participant: NewParticipant, jwt_token: str | None = No
...
+            )
+
+            new_participant_object_id = str(
+                new_participant.inserted_id,
+            )
+            # Adds the participant to the team
+            available_team = TeamFunctionality.add_participant_to_team_object(
+                available_team.team_name, new_participant_object_id,
+            )
+            cls.update_participant(
```
IvanObreshkov, line 269: Should this be ParticipantsFunctionality.update_participant?

packages/py-api/py_api/controllers/hackathon/participants_controller.py
```diff
@@ -240,64 +241,70 @@ def add_participant(cls, participant: NewParticipant, jwt_token: str | None = No
...
+            new_participant_object_id = str(
+                new_participant.inserted_id,
+            )
+            # Create New Team and assign the new participant to it
+            newTeam = TeamFunctionality.create_team_object_with_admin(
+                user_id=new_participant_object_id, team_name=TeamFunctionality.generate_random_team_name(), generate_random_team=True,
+            )
+
+            if newTeam:
+                cls.update_participant(
```
IvanObreshkov, line 299: This one as well? (ParticipantsFunctionality.update_participant)

packages/py-api/py_api/controllers/hackathon/participants_controller.py
```diff
@@ -240,64 +241,70 @@ def add_participant(cls, participant: NewParticipant, jwt_token: str | None = No
...
+                    new_participant_object_id, UpdateParticipant(
+                        team_name=newTeam.team_name,
+                    ),
+                )
+                TeamFunctionality.insert_team(newTeam)
```
IvanObreshkov, line 304: new line

IvanObreshkov (commented):
Resolve the comments, otherwise BRAVO!

packages/py-api/py_api/controllers/hackathon/participants_controller.py
Isak-Bego: Which check?

packages/py-api/py_api/controllers/hackathon/participants_controller.py
Isak-Bego: Also isnt this the same as the update_participant we have on the participantsController?

IvanObreshkov (approved):
LGTM! Again great work!

Issue #636, opened 2024-02-13
Fix HackAUBG navigation centering


Aleks1908:
## Brief description:
The navigation on the HackAUBG page is not always centered

PR #637, opened 2024-02-13
Fix navigation centering in HackAUBG page 

Aleks1908:
#636 
Added filler div to keep navigation centered


IvanObreshkov (approved):
LGTM!

PR #638, opened 2024-02-15
#635 Adds new fields to the pydantic participant model

### LINKED ISSUE #635: Add new fields to the Participant Pydantic Model
## Brief description:
Currently, we have these:

```python
    first_name: str
    last_name: str
    email: EmailStr
    tshirt_size: TSHIRT_SIZE
```
And the TSHIRT_SIZE is defined like this:
```python
TSHIRT_SIZE = Literal["S", "M", "L"]
```
What fields do we need:
 * "XL" to be added in the TSHIRT_SIZE
 * `university` (required) Options: "Sofia University", "Technical University - Sofia", "Plovdiv University", "Other"
 * `location`  str (required)
 * `age` int with constraints  (required) (it cannot be a negative number, 0 and it should not be bigger than 99 or 120)
 * `source_of_referral` (optional) How did you find out about HackAUBG: "University", "Friends", "I was on a previous edition of Hack AUBG"
  * `programming_language` (optional) What programming languages are you familiar with: "Frontend Programming", "Backend Programming", "Programming in C#", "Programming in Java", "Programming in Python", "Programming in JavaScript", "Other"
  * `programming_level` (optional) What is your programming level: "Beginner", "Intermediate", "Advanced", "I am not participating as a programmer", "Other"
  * `has_participated_in_hackaubg`: bool (required)
  * `has_internship_interest`: bool (required)
  * `has_participated_in_hackathons`: bool (required)
  * `has_previous_coding_experience`:  bool (required)
  * `newsletter_consent`:  bool (required)
  * `share_info_with_sponsors`:  bool (required) (This should always be True)

## How to achieve it:
* Add these fields to the two [models](https://github.com/AUBGTheHUB/monolith/blob/master/packages/py-api/py_api/models/hackathon_participants_models.py)
* For the multiple choice you can again use Literal. But make sure to check with @Aleks4o about how the options are written as they have to be the same on the front end to avoid Pydantic crying about wrong inputs.
* [React Registration Form](https://github.com/AUBGTheHUB/monolith/blob/52bd7d3acaa6d2e0b14f9455b232db01d9ae9ea2/packages/web/src/components/spa/HackAUBG/RegistrationForm/RegistrationForm.jsx) from last year if you are interested

If you have questions, don't hesitate to ask @IvanObreshkov or @Aleks1908. Good luck!

Isak-Bego:
## Changes:
Added the new fields to the Pydantic model along with the respective validation like specified.
Changed the update_participant() function from the participant_base file so that it can update booleans from false->true and vice-versa. 
Changed the return statement so that it returns the updated participant because previously it was returning the before changes. 

resolves #635 

## How to verify:

- [ ] Start the Backend.

- [ ] Send a POST request to the participant endpoint to create a new Participant 

- [ ] Play around with the body of the request and check the responses

- [ ] Send a PUT request to the participant endpoint along with the object id of the targeted participant

- [ ] Play around with the body and check the responses

### DIFF
--- a/packages/py-api/py_api/functionality/hackathon/participants_base.py
+++ b/packages/py-api/py_api/functionality/hackathon/participants_base.py
@@ -35,7 +35,11 @@ class ParticipantsFunctionality:
     def update_participant(cls, object_id: str, participant: UpdateParticipant | NewParticipant) -> Any:
         obj = {
             key: value for key, value in participant.model_dump().items() if
-            value
+            value != None
         }
 
-        return cls.pcol.find_one_and_update({"_id": ObjectId(object_id)}, {"$set": obj})
+        cls.pcol.find_one_and_update(
+            {"_id": ObjectId(object_id)}, {"$set": obj},
+        )
+        # Returns the updated participant
+        return cls.pcol.find_one({"_id": ObjectId(object_id)})
--- a/packages/py-api/py_api/models/hackathon_participants_models.py
+++ b/packages/py-api/py_api/models/hackathon_participants_models.py
@@ -1,8 +1,25 @@
-from typing import Literal, Optional
+from typing import Annotated, Literal, Optional
 
-from pydantic import BaseModel, EmailStr
+from pydantic import BaseModel, EmailStr, Field, validator
 
-TSHIRT_SIZE = Literal["S", "M", "L"]
+TSHIRT_SIZE = Literal["S", "M", "L", "XL"]
+UNIVERSITIES_LIST = Literal[
+    "Sofia University",
+    "Technical University - Sofia", "Plovdiv University", "Other",
+]
+ALLOWED_AGE = Annotated[int, Field(ge=0, le=99)]
+REFERRAL_SOURCES_LIST = Literal[
+    "University",
+    "Friends", "I was on a previous edition of Hack AUBG",
+]
+PROGRAMMING_LANGUAGES_LIST = Literal[
+    "Frontend Programming", "Backend Programming", "Programming in C#",
+    "Programming in Java", "Programming in Python", "Programming in JavaScript", "Other",
+]
+PROGRAMMING_LEVELS_LIST = Literal[
+    "Beginner", "Intermediate",
+    "Advanced", "I am not participating as a programmer", "Other",
+]
 
 
 class NewParticipant(BaseModel):
@@ -13,6 +30,24 @@ class NewParticipant(BaseModel):
     team_name: Optional[str] = None
     is_verified: Optional[bool] = False
     is_admin: Optional[bool] = False
+    university: UNIVERSITIES_LIST
+    location: str
+    age: ALLOWED_AGE
+    source_of_referral: Optional[REFERRAL_SOURCES_LIST] = None
+    programming_language: Optional[PROGRAMMING_LANGUAGES_LIST] = None
+    programming_level: Optional[PROGRAMMING_LEVELS_LIST] = None
+    has_participated_in_hackaubg: bool
+    has_internship_interest: bool
+    has_participated_in_hackathons: bool
+    has_previous_coding_experience: bool
+    newsletter_consent: bool
+    share_info_with_sponsors: bool
+
+    @validator("share_info_with_sponsors")
+    def check_if_true(cls, value: bool) -> bool:
+        if not value:
+            raise ValueError("share_info_with_sponsors shall be true")
+        return value
 
 
 class UpdateParticipant(BaseModel):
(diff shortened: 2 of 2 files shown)


packages/py-api/py_api/models/hackathon_participants_models.py
```diff
@@ -21,4 +56,16 @@ class UpdateParticipant(BaseModel):
     email: Optional[EmailStr] = None
     tshirt_size: Optional[TSHIRT_SIZE] = None
     team_name: Optional[str] = None
-    is_verified: Optional[bool] = False
+    is_verified: Optional[bool] = None
```
IvanObreshkov, line 59: Why this is None? It should be False.

packages/py-api/py_api/models/hackathon_participants_models.py
```diff
@@ -1,8 +1,25 @@
...
-from pydantic import BaseModel, EmailStr
+from pydantic import BaseModel, EmailStr, Field, validator
 
-TSHIRT_SIZE = Literal["S", "M", "L"]
+TSHIRT_SIZE = Literal["S", "M", "L", "XL"]
+UNIVERSITIES_LIST = Literal[
+    "Sofia University",
+    "Technical University - Sofia", "Plovdiv University", "Other",
+]
+ALLOWED_AGE = Annotated[int, Field(ge=0, le=99)]
```
IvanObreshkov, line 10: Do you believe we should have newborns as participants?😂

packages/py-api/py_api/functionality/hackathon/participants_base.py
```diff
@@ -35,7 +35,11 @@ def delete_participant(cls, object_id: str) -> results.DeleteResult | None:
     def update_participant(cls, object_id: str, participant: UpdateParticipant | NewParticipant) -> Any:
         obj = {
             key: value for key, value in participant.model_dump().items() if
-            value
+            value != None
```
IvanObreshkov, line 38: This is not needed. "If value" does the same as value != None, it's just more concise. In Python None, False, Empty List, Empty Dict, Empty Set and so on are falsy values.

packages/py-api/py_api/models/hackathon_participants_models.py
```diff
@@ -1,8 +1,25 @@
...
+REFERRAL_SOURCES_LIST = Literal[
+    "University",
+    "Friends", "I was on a previous edition of Hack AUBG",
```
IvanObreshkov, line 13: Please put new lines between the constants as they are a bit long. This is for every constant.

packages/py-api/py_api/models/hackathon_participants_models.py
```diff
@@ -13,6 +30,24 @@ class NewParticipant(BaseModel):
...
+    programming_language: Optional[PROGRAMMING_LANGUAGES_LIST] = None
+    programming_level: Optional[PROGRAMMING_LEVELS_LIST] = None
+    has_participated_in_hackaubg: bool
+    has_internship_interest: bool
+    has_participated_in_hackathons: bool
+    has_previous_coding_experience: bool
+    newsletter_consent: bool
+    share_info_with_sponsors: bool
+
+    @validator("share_info_with_sponsors")
```
IvanObreshkov, line 46: Bravo!

IvanObreshkov (commented):
Looks good, a couple of nits and we should be good to go. Also nice on using the validator, this is the proper way!

packages/py-api/py_api/functionality/hackathon/participants_base.py
asynchroza: Damn, what is this dude talking about?

IvanObreshkov: A Микаел Again.

packages/py-api/py_api/models/hackathon_participants_models.py
Isak-Bego: It wasnt specified on the ticket. Shall I set it above 18?

packages/py-api/py_api/models/hackathon_participants_models.py
Isak-Bego: To implement the functionality for updating boolean values. If you want to change a bool field back to false you cannot do it if it isnt None. That is why I added the check != None.

packages/py-api/py_api/functionality/hackathon/participants_base.py
Isak-Bego: > Hmmm. Are you sure about the logic here? We might want to replace a field with an empty list which in your case will be skipped. I think Isak's explicit check is the right way to go.

This is exactly why I hvae done it this way. Because that was the case if you want to set a boolean field back to false, it would ignore it.

packages/py-api/py_api/models/hackathon_participants_models.py
Isak-Bego: Thank you man!

IvanObreshkov (approved):
Again. Job done at a high standart!

PR #642, opened 2024-02-17
Update urls and check for env

IvanObreshkov:
Changes made:
* Updated urls for frontend
* Update logic for domain depending on evn


Aleks1908 (approved):
lgtm

PR #647, opened 2024-02-17
Refactor exception handling and mongo queires

IvanObreshkov:
## Changes made:
* Refactored Mongo queries with transactions
* Optimized exception handling and error messages


asynchroza (approved):
I believe in you

Issue #602, opened 2024-01-08
Registration Component


Aleks1908:
## Brief description:
* Implement Registration by using the React Hook Form library
* Implement input validation
* Make registration only available if the total amount of teams is not reached
* Make it flexible for all Registration cases. Team creation (team field required), Solo register(no team field) and register for existing team (no team field but team name defined in jwt token)

What fields do we need:
Inputs:
 * Full Name (required)
 * Email (required)
 * Age (required)
 * Location (required)
* What are your job interests(required)
* Do you have a team (required)
* If team -> What is the name of your team?(required)

Dropdowns:
 * Choose an School/University (optional): "AUBG", "Sofia University", "Technical University - Sofia", "Plovdiv University", "Other"
 * T-shirt size (optional): "Small (S)", "Medium (M)", "Large (L)", "Extra Large (XL)"
 * How did you find out about HackAUBG? (required): "University", "Friends", "I was on a previous edition of Hack AUBG", "other"
 *  What programming languages are you familiar with?(required): "Frontend Programming", "Backend Programming", "Programming in C#", "Programming in Java", "Programming in Python", "Programming in Javascript"
 * What is your programming level? (required): "Beginner", "Intermediate", "Advanced", "I am not participating as a programmer", "Other"

Yes/No Questions (all are required):
* Have you participated in Hack AUBG before?
* Are you looking for an internship?
* Have you participated in other Hackathons?
* Do you have previous coding experience?
* Do you want to receive our newsletter with potential job offerings?

Checkbox:
* Agreement to share information with sponsors (must be selected)

PR #652, opened 2024-02-22
#602 Registration component and small fixes to navigation hover colors

Aleks1908:
## Remade the Registration component 
Added components for each type of field and reworked the logic of the Registration form. And some small tweaks on the Navigation. Completed with the help of @Pastedko 

Resolves: #602 

## How to verify:

- [ ] Open the RegistrationForm.jsx and remove the check for the feature switches

- [ ] Run the Front-end and the py-api and register an account. 

- [ ] You should receive an email with verification (that will not work as of the moment because the verification page is not merged)


packages/web/src/components/spa/HackAUBG/RegistrationForm/InputComponent.jsx
r-mishev, line 1: This component can be refactored to not use this many nested if statements.

packages/web/src/components/spa/HackAUBG/RegistrationForm/RegistrationForm.jsx
```diff
@@ -40,640 +41,298 @@ const RegistrationForm = () => {
...
-            if (field[1] === 'True' || field[1] === 'False') {
-                data[field[0]] = field[1] === 'True' ? true : false;
+            for (const [key, value] of Object.entries(values)) {
+                if (value === 'true') {
+                    updatedValues[key] = true;
+                } else if (value === 'false') {
+                    updatedValues[key] = false;
+                } else {
+                    updatedValues[key] = value;
+                }
```
r-mishev, line 65: Why not just

```suggestion
updatedValues[key] = value
```

packages/web/src/components/spa/HackAUBG/RegistrationForm/RegistrationForm.jsx
```diff
@@ -40,640 +41,298 @@ const RegistrationForm = () => {
...
-        setSubmitButtonValue('Submit');
     };
 
-    useEffect(checkButtonAvailability, [Object.keys(errors)]);
-
-    const showButton = () => {
-        if (loadingAnimation) {
-            return <input type="submit" value="LOADING" className="hackaubg-register-btn" />;
+    const display = () => {
+        if (getValues().team == 'true') {
```
r-mishev, line 88: Are you certain it shouldn't be `` === `` ?

r-mishev (approved):

packages/web/src/components/spa/HackAUBG/VerifyAccountPop/VerifyAccount.jsx
```diff
@@ -1,14 +1,23 @@
...
+    const [jwtToken, setJwtToken] = useState('');
     const location = useLocation();
-    const params = new URLSearchParams(location.search);
-    const jwtToken = params.get('jwt_token');
+    const [buttonState, setButtonState] = useState(true);
+
+    useEffect(() => {
+        const params = new URLSearchParams(location.search);
+        if (params.size === 0) {
+            window.location.href = '/hackaubg';
```
asynchroza, line 16: Isn't the url already that? No need to change the value.

packages/web/src/components/spa/HackAUBG/VerifyAccountPop/VerifyAccount.jsx
```diff
@@ -1,14 +1,23 @@
...
+        } else {
+            setJwtToken(params.get('jwt_token'));
+        }
+    }, [location]);
```
asynchroza, line 20: I would usually suggest extracting those as hooks but at this point just make it work.

packages/web/src/components/spa/HackAUBG/VerifyAccountPop/VerifyAccount.jsx
Aleks1908: This is a check if after verify there is a jwt_token. And if there is not and the url looks like this /hackaubg/verify it goes back to /hackaubg

asynchroza: okay, makes sense

PR #669, opened 2024-03-03
662 Implements sending successful registration messages to participants

### LINKED ISSUE #662: Implement sending of successful registration emails
## Brief description:
* After a random participant has verified their email we should send them a second email telling them they have successfully registered for the hackathon.
* We should do the same for the admin of a team, before sending them the email with an invitation link.
* After a participant has registered via the invite link we should again send them an email they have successfully registered for the hackathon.
## How to achieve it:
Here is an example of how seeing the invite link is done
https://github.com/AUBGTheHUB/monolith/blob/5ea564cda500a73438e423bec4313358f10dedea/packages/py-api/py_api/controllers/hackathon/verification_controller.py#L55-L60

Your code should be in the verification_controller.

Isak-Bego:
## Implemented sending a successful registration message to the participants:
I added a background task in the verification controller to send an email to the verified random participant and modified the message sent to the admin of a normal team to also include a successful registration message. Furthermore, for the participants who are registered by link, the logic for sending the email is implemented in the participant's controller in the method named add_participant_to_existing_team(), because they are verified by default and they dont use the verify endpoint. I also changed some confusing logic about how the check for the capacity reach was made in the add_participant method to something that I think is more clear and correct. The previous version wasn't letting you add a participant to an existing team when the number of teams had reached the cap. This was because it had a check in the beginning of the file which returned a json response when the cap was reached with causing issues to the logic of the code. 
Lastly, I changed the method get_count_of_teams to only count the verified teams in the database.

resolves #662  

## How to verify:

- [ ] Start the backend

- [ ] Register as a participant with a team, as a random participant and as a participant with invitation link. 

- [ ] Check the emails that you have received.

### DIFF
--- a/packages/py-api/py_api/controllers/hackathon/participants_controller.py
+++ b/packages/py-api/py_api/controllers/hackathon/participants_controller.py
@@ -118,9 +118,6 @@ class ParticipantsController:
         str, str,
     ]:
 
-        if TeamFunctionality.get_count_of_teams() > 16:
-            return JSONResponse(content={"message": "Hackathon is at max capacity"}, status_code=409)
-
         if ParticipantsFunctionality.check_if_email_exists(participant.email):
             return JSONResponse(status_code=400, content={"message": "User with such email already exists"})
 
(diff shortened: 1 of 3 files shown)


IvanObreshkov (approved):
I believe in you!

PR #676, opened 2024-03-08
#640 create email templates and remove newsletter from registration

Aleks1908:
## Added React Email Library and Created Email templates
- Added new gum option to launch the emails service
- Created new email templates
- Remove unused field from registration


DemonHashira (approved):
Top G

PR #694, opened 2024-03-19
Update schedule

Aleks1908:
Changed schedule section text and made university filed required because it is required in the backend


DemonHashira (approved):
🤜🤛🤝

PR #702, opened 2024-04-05
Fix undef length

Aleks1908:
In the sponsors component added check for is api loading and rendering the items after they load


IvanObreshkov (approved):
staa da

PR #716, opened 2024-09-20
#711 create repository service layer

### LINKED ISSUE #711: Create the DB Layer
## Brief description:
* Create a proper abstraction over the pymongo lib
* Create a CRUDRepository interface

IvanObreshkov:
Resolves #711 #717

* Added an abstraction over motor lib
* Added CRUDRepository interface and impl
* Added a model and service layers


services/py_api/src/database/query_manager.py
```diff
@@ -0,0 +1,103 @@
...
+            await session.end_session()
+
+    async def fetch_by_id(self) -> Result:
+        pass
+
+    async def fetch_all(self) -> Result:
+        # TODO: Finish. Use the self._collection if needed
+        pass
+
+    async def update_obj_tx(self) -> Result:
```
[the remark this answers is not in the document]
asynchroza: Something I do is have an additional kwarg `opts` where if you pass the transaction, it will use it instead of using the db client.

services/py_api/src/database/repositories/participants_repository.py
```diff
@@ -0,0 +1,43 @@
...
+        if is_err(result):
+            return result
+
+        return Ok(
+            ParticipantResponse(
+                id=str(result.value.inserted_id),
+                name=input_data.name,
+                email=input_data.email,
+                is_admin=True,
+                created_at=datetime.now(),
```
asynchroza, line 40: These timestamps will be inconsistent with the actual data.

services/py_api/src/database/query_manager.py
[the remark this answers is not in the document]
IvanObreshkov: > Transactions add overhead. What if it's a single update? You don't need a transaction for this.

I prefer to do write operations in a transactional context. By doing so, we can do a "request scoped" operation. Also, if there is an error, we don't have to worry about permanent changes to the database. I suggest we add a method which does the update and write without a transaction and the caller can choose.

services/py_api/src/database/repositories/participants_repository.py
IvanObreshkov: > These timestamps will be inconsistent with the actual data.

I know. I will fix it

services/py_api/src/database/query_manager.py
```diff
@@ -0,0 +1,103 @@
...
+from structlog.stdlib import get_logger
+
+from src.database.db_manager import DatabaseManager
+
+LOG = get_logger()
+
+T = TypeVar("T")
+
+
+class QueryManager:
```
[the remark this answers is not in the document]
IvanObreshkov: > Do you really have to manage all of this yourself? Doesn't pymongo provide this out of the box?

it provides a method A[gnosticClientSession.with_transaction()](https://motor.readthedocs.io/en/stable/api-tornado/motor_client_session.html#motor.motor_tornado.MotorClientSession.with_transaction). But I don't want to have a timeout of 120 sec. Also, I prefer to have more control.

services/py_api/src/database/query_manager.py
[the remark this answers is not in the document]
asynchroza: Yes, I'm not saying ditch transactions. I'm just saying that there's no need to initialize a transaction when you're doing just a single write operation. I don't know how it works in the case of pymongo but I'm using Drizzle now and tx and db are interchangable. Hence, in my case, it doesn't make sense to have two separate methods but rather just one with optional tx param.

services/py_api/src/database/query_manager.py
```diff
@@ -0,0 +1,103 @@
...
+
+        except Exception as e:
+            LOG.exception("Aborting transaction due to err {}".format(e))
+            await session.abort_transaction()
+            return Err("Unexpected err occurred during transaction")
+
+        finally:
+            LOG.debug("Closing DB session")
+            # Finish this session. If a transaction has started, abort it.
+            await session.end_session()
```
asynchroza, line 88: Wait, so you're creating a transaction for each operation? 
So, if you have three write operations, these are three totally separate transactions?
Am I understanding this correctly?

asynchroza: If that's the case, you should propagate the transactions from a level below. These methods shouldn't be responsible for creating the sessions/transactions themselves. This should happen at the repository level.

asynchroza: Otherwise, you won't be able to chain the operations together

services/py_api/src/database/query_manager.py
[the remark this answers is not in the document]
IvanObreshkov: It means that a transaction is done specifically for this request. If we have an error somewhere in the request we should about the transaction. Hense you don't have to do a clean_up.

services/py_api/src/database/query_manager.py
IvanObreshkov: Yeah i am not sure if i have done it properly as it was 5am. I will take a look again.

services/py_api/src/database/query_manager.py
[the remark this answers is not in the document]
asynchroza: That's usually how transactions are supposed to work. In your case, you would want to first create the transaction, then wrap all chained operations in a try catch clause and abort if something errors out.

Isak-Bego (commented):
Overall, seems well written. I only have that one comment about the deadline check. I believe you also mentioned that you plan to make some changes. Thank you Ivan! P.S The repository architecture is nice : )

IvanObreshkov:
@asynchroza @Isak-Bego I made a lot of changes, pls review them again. Also if everything is ok @Isak-Bego pls fix the tests :D

services/py_api/src/database/query_manager.py
```diff
@@ -0,0 +1,103 @@
...
+
+    _RETRY_TRANSACTION_DEADLINE_SEC = 3
+
+    def __init__(self, db_manager: DatabaseManager, collection: str) -> None:
+        self._client = db_manager.client
+        self._collection = self._client[db_manager.DB_NAME][collection]
+
+    def _deadline_exceeded(self, start_time: float) -> bool:
+        return monotonic() - start_time < self._RETRY_TRANSACTION_DEADLINE_SEC
+
```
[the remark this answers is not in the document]
IvanObreshkov: @Isak-Bego Fixed in b7cca32bcf584105b0b1c1b3f2e490a9c8299cca

services/py-api/src/database/db_manager.py
```diff
@@ -0,0 +1,88 @@
...
+    DB_NAME = "TheHubDEV"
+
+    def __init__(self) -> None:
+        # The mongo client has a conn pool under the hood, and we set a min number of idle conn that the pool has
+        # to maintain, the default is 0. After maxIdleTimeMS the connection pool replaces the idle conn with a new one.
+        # I set a min number of conn according to the article below. The connection is also removed too fast, after
+        # around 30 sec, that's why I increased maxIdleTimeMS to 5 min.
+        # https://alexedwards.net/blog/configuring-sqldb
+        # https://www.mongodb.com/docs/languages/python/pymongo-driver/current/faq/#how-does-connection-pooling-work-in-pymongo-
+        self._client = AsyncIOMotorClient(host=os.environ["DATABASE_URL"], minPoolSize=25, maxIdleTimeMS=5 * 60 * 1000)
```
asynchroza, line 30: Why would you have such a high `minPoolSize`? This is processing power which you're wasting on managing open connections especially when you know that you don't have such high traffic.

IvanObreshkov: I got it from this article: https://alexedwards.net/blog/configuring-sqldb
What number do you suggest?

asynchroza: I skimmed and found no reference of min value there

asynchroza: I suggest you let the client handle itself.

IvanObreshkov: We could set the minPoolSize to be 10% of the maxPoolSize: https://medium.com/@dhanushkasampath.mtr/what-are-the-default-values-for-hikari-connection-pool-if-we-do-not-override-in-application-properti-11932cdbe321

IvanObreshkov:
> I will probably have some time to review it tomorrow but an ask for the future is to have your work broken down into smaller pieces as a PR with 54 files changed is a pretty substantial amount of changes to read through.

They are around 15 but I changed the name of the folder from `py_api` to `py-api` as you suggested, that's why it says 54.

Isak-Bego (commented):
It looks really good. I like the fact that it provides several layers of abstraction for handling the requests. Elegant. Just a couple of formating/typo issues. Thank you Ivan!

services/py-api/src/database/repository/base_repository.py
```diff
@@ -0,0 +1,38 @@
...
+        raise NotImplementedError()
+
+    @abstractmethod
+    async def fetch_all(self) -> Result:
+        raise NotImplementedError()
+
+    @abstractmethod
+    async def update(
+        self, obj_id: str, input_data: BaseModel, session: Optional[AsyncIOMotorClientSession] = None
+    ) -> Result:
```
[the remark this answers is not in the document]
asynchroza: That's the formatting rules pretty sure

services/py-api/src/database/db_manager.py
asynchroza: It's up to you. As mentioned I would usually just have the defaults until proven ineffective.

services/py-api/src/database/db_manager.py
```diff
@@ -0,0 +1,88 @@
...
+
+        LOG.debug("Pong")
+
+    @property
+    def client(self) -> AsyncIOMotorClient:
+        return self._client
+
+
+def ping_db() -> Err[str]:
+    """This method is used on application startup. It is not async as I cannot await the start() method"""
```
asynchroza, line 57: This comment is confusing. Why do you need an async method? 

It's a bit strange you're adding a docstring comment and talking in first person. If this is not a comment which is to be exposed by the lsp utilities, have it as a regular comment.

services/py-api/src/database/repository/participants_repository.py
```diff
@@ -0,0 +1,54 @@
...
+from src.server.exception import DuplicateEmail
+from src.server.schemas.request_schemas.schemas import ParticipantRequestBody
+
+LOG = get_logger()
+
+
+class ParticipantsRepository(CRUDRepository):
+    # TODO: Implement the rest of the methods
+    def __init__(self, db_manager: DatabaseManager, collection: str) -> None:
+        self._client = db_manager.client
```
asynchroza, line 22: Probably a good idea is to have a `get_client` and `get_collection` methods and define the client as a private member of the `db_manager`

services/py-api/src/database/repository/participants_repository.py
```diff
@@ -0,0 +1,54 @@
...
+    async def update(
+        self, obj_id: str, input_data: BaseModel, session: Optional[AsyncIOMotorClientSession] = None
+    ) -> Result:
+        raise NotImplementedError()
+
+    async def delete(self, obj_id: str, session: Optional[AsyncIOMotorClientSession] = None) -> Result:
+        raise NotImplementedError()
+
+    async def create(
+        self, input_data: ParticipantRequestBody, session: Optional[AsyncIOMotorClientSession] = None
```
asynchroza, line 40: What's the whole deal with async in the first place?

services/py-api/src/database/repository/participants_repository.py
IvanObreshkov: It is defined as a private member?

services/py-api/src/database/repository/participants_repository.py
```diff
@@ -0,0 +1,54 @@
...
+    ) -> Ok[Participant] | Err[Type[DuplicateEmail]] | Err[Exception]:
```
asynchroza, line 41: That's a weird return type. You're forcing yourself into mandatory type check if value is of type `Err`.

services/py-api/src/database/repository/participants_repository.py
IvanObreshkov: As we have async handling of requests it makes sense to use an async library, in this case `motor`. Otherwise the sync operations of the normal `pymongo` library would block the event loop.

services/py-api/src/database/repository/teams_repository.py
```diff
@@ -0,0 +1,54 @@
...
+
+
+class TeamsRepository(CRUDRepository):
+    def __init__(self, db_manager: DatabaseManager, collection: str) -> None:
+        self._client = db_manager.client
+        self._collection = self._client[db_manager.DB_NAME][collection]
+
+    async def create(
+        self, input_data: ParticipantRequestBody, session: Optional[AsyncIOMotorClientSession] = None
+    ) -> Ok[Team] | Err[DuplicateKeyError] | Err[Exception]:
```
asynchroza, line 26: `Err[DuplicateKeyError | Exception]`

services/py-api/src/database/repository/participants_repository.py
IvanObreshkov: This was because of mypy i think.

services/py-api/src/database/repository/teams_repository.py
asynchroza: Or better yet, define a base error which you can use as the base type and return errors which extend it

services/py-api/src/database/repository/teams_repository.py
```diff
@@ -0,0 +1,54 @@
...
+
+        try:
+            team = Team(_id=ObjectId(), name=input_data.team_name)
+            LOG.debug("Inserting team {}".format(team.model_dump()))
+            await self._collection.insert_one(document=team.__dict__, session=session)
+            return Ok(team)
+
+        except DuplicateKeyError:
+            LOG.warning("Team insertion failed due to duplicate team name {}".format(input_data.team_name))
+            return Err(DuplicateTeamName("Duplicate team name"))
```
asynchroza, line 36: `DuplicateTeamNameError` and pass just the team name

services/py-api/src/database/transaction_manager.py
```diff
@@ -0,0 +1,87 @@
...
+class TransactionManager:
+    _RETRY_TRANSACTION_DEADLINE_SEC = 3
+
+    def __init__(self, db_manager: DatabaseManager) -> None:
+        self._client = db_manager.client
+
+    def _deadline_exceeded(self, start_time: float) -> bool:
+        return monotonic() - start_time > self._RETRY_TRANSACTION_DEADLINE_SEC
+
+    async def _retry_tx(self, func: Callable[..., Awaitable[Result[T, E]]], *args: Any, **kwargs: Any) -> Result[T, E]:
```
asynchroza, line 28: I would suggest that you write a good documentation around this class as it might be pretty confusing to a newcomer especially the types.

services/py-api/src/database/repository/participants_repository.py
asynchroza: I'm talking about the `db_manager.client` and the collection fetch you do the line below

services/py-api/src/database/repository/participants_repository.py
asynchroza: What async handling do you have?

IvanObreshkov: FastAPI as a whole is async. I can't understand your question.

services/py-api/src/server/routes/participants_routes.py
```diff
@@ -0,0 +1,49 @@
...
+from src.service.participants_service import ParticipantService
+
+participants_router = APIRouter(prefix="/hackathon/participants")
+
+
+# https://fastapi.tiangolo.com/tutorial/dependencies/sub-dependencies/
+# Dependency wiring
+
+
+def _p_repo(db_manager: DB_MANAGER) -> ParticipantsRepository:
```
asynchroza, line 20: Help me understand why you need these methods here.

IvanObreshkov: There is a comment pointing to the docs. This is how Dependency injection is done in FastAPI. I know it's weird, I was also confused the first time.

services/py-api/src/database/repository/participants_repository.py
asynchroza: Yes. FastAPI is built around async but doesn't force you into using async. Asynchronous code is resource demanding and doesn't make sense to use it if you don't actually need it.

asynchroza: Also, async in this context just adds an additional layer of complexity.

asynchroza: The only functionality I can think of that might possibly need an async operation would be the functionality around sending emails.

services/py-api/src/server/routes/participants_routes.py
asynchroza: Okay. Got it.

services/py-api/src/database/repository/participants_repository.py
IvanObreshkov: What about db queries? These are also IO bound operations. In JS I have seen db libraries which are async.

asynchroza: Yes. Async in JavaScript is a native concept. But let's first take a step back and figure out what async really means. Async is like running operations in parallel, right? Why do you need to overwhelm your code with async resolutions and event loop processing if your code doesn't have to run multiple independent of each other operations at once? 

Think of the following case. You want to create 5 team members at once and add all of them to a team. This makes sense to be an async operation. You batch these 5 queries and execute them in "parallel" context, when all of these are resolved, take the ids and insert them into team entity. 

In your case, I'm not saying to totally get rid of async but instead bring it's usage to a minimum if you don't really need it. I can't think of operations in your business use cases where you would need such an async operation where you want to keep running the thread without blocking it.

IvanObreshkov: @asynchroza Ok, from my knowledge and [this MDN article ](https://developer.mozilla.org/en-US/docs/Learn/JavaScript/Asynchronous/Introducing) async is primarily used when you have IO-bound operations, which could take some time. In our case, a db query could take some time, but we don't want to wait for the query to finish (block the thread) before accepting the next client request. 

Enough about theory, the best way to see which approach is better is to load test the API, so that's what we did with @Isak-Bego. 
The tests were performed on Isak's MacBook Pro (M3 pro, not sure about the concrete one and 18 GB of RAM)

We used a single `uvicorn` worker and we left the config of the connection pool with a minimum of 25 idle connections. Our MongoDB instance is deployed in AWS Frankfurt (eu-central-1).

For the load test, we used `locust` with a max concurrency (number of users) of 200 and a ramp-up rate of 10. Each test was performed 3 times.  Each test was performed on a clean db (0 participants and 0 teams).

The 2 screenshots below are from each group of tests, and the numbers I have written are the average from the three tests performed for each test case.

**The first test case was performed with the current state of the code (full async and `motor` lib):**


image
image

Average response time in ms from the three tests: 3593.29
Average number of failed requests from the three tests: 0.33 (We had only one failed request for all three tests):
Cause of failure:

image

-----------------------------------------------------------------------------------------------------------------

**The second test case was performed with no async anywhere using the `pymongo` library.**


image
image

Average response time in ms from the three tests: 4037.56
Average number of failed requests from the three tests: 1.33 (We had a total of  failed 4 requests for all tests):
Cause of failure:

image

-----------------------------------------------------------------------------------------------------------------

**BONUS: The third test case was performed with async only on the route using the `pymongo` library. The test was performed only once with a max concurrency of 100 users and a ramp-up of 10**
image

Here, we wanted to see if the event loop would really get blocked (it did), as we are executing a synchronous operation in an async context.

image
image

-----------------------------------------------------------------------------------------------------------------

Summary: 
**We are 444,27 ms faster on average in terms of response time, when we use the async Mongo library 

**`motor`.** And we have a lower number of failed requests, though when taking the average, it's only 1.**

asynchroza: Thank you for taking the time to analyze this and compile the report, it's amazing and proves I was mistaken.

Upon revisiting my original comment, I realize I was primarily focused on how async affects a single method and its I/O dependency, rather than considering the broader global context. As we discussed in the chat, the choice here seems to be between prioritizing a highly efficient API or one that’s easier for newcomers to adopt.

IvanObreshkov: Thank you Misho, means a lot from you!

services/py-api/src/database/db_manager.py
IvanObreshkov: Fixed in [17e3ee7](https://github.com/AUBGTheHUB/monolith/pull/716/commits/17e3ee70284456f6830134d48e3e1855c1afd301)

services/py-api/src/database/repository/participants_repository.py
IvanObreshkov: The db client is a private member of the DatabaseManager and`db_manager.client` is defined as a property (getter):
image

image

Should I rename it to get_client?

The `self._collection = self._client[db_manager.DB_NAME][collection]` on the next line is required as every repository is responsible for working with a particular collection, and in order for us to create an ` AsyncIOMotorCollection` we need the client. I can do it like this for simplicity:
```python
self._collection = db_manager.client[db_manager.DB_NAME][collection]
```
I am not sure we need a getter for the collection itself.

Aleks1908 (changes requested):
Make reviews instead of single line comments dumbass

services/py-api/src/database/repository/teams_repository.py
IvanObreshkov: Fixed in 1a00dd8c6f9016c7770fd6e17ffc6b8195dc8ce2

services/py-api/src/database/repository/teams_repository.py
IvanObreshkov: Fixed in 1a00dd8c6f9016c7770fd6e17ffc6b8195dc8ce2

services/py-api/src/database/repository/participants_repository.py
IvanObreshkov: Fixed in 1a00dd8c6f9016c7770fd6e17ffc6b8195dc8ce2

services/py-api/src/database/repository/participants_repository.py
IvanObreshkov: Fixed in 35cfe383f588e86496b7f3e3fb8e855965ebe35a

Now it is a better interface indeed, thank you!

services/py-api/src/database/db_manager.py
IvanObreshkov: Made some changes in [c2f5f08](https://github.com/AUBGTheHUB/monolith/pull/716/commits/c2f5f08237741c90057394088566612c86c3f99c). Let's leave it like this, later I will make some performance testing to see the differences.

services/py-api/src/database/transaction_manager.py
IvanObreshkov: Fixed in 3dbcb9b202de882ccd1fe0025df48c3c6fff976c

Isak-Bego (approved):
LGTM

PR #718, opened 2024-09-23
#715 auto install py api

### LINKED ISSUE #715: Add automatic installation of py-api
## Brief description:
We need a `Makefile` in the root again. We can use `gum` again. I would advise you to check the meaning of `.PHONEY` and decide if we need it. We can have different make targets for installing the different services and running them.
## How to achieve it:
I would recommend creating an `install.sh` script which should do the following:
1. Check the current ENV (Linux or MAC)
	1. If we are on MAC: check if `brew` is installed, if it's not install it
2. Install python
	1.  On Mac do `brew install python3.12.6`
	2. On Linux use `apt` 
	3. A note about `pyenv`:
	We could also use `pyenv` but you should check the difference between having the `brew/apt python3.12.6` as the default python version on the system or using `pyenv install python3.12.6` and [`pyenv global 3.12.6`](https://github.com/pyenv/pyenv/blob/master/COMMANDS.md#pyenv-global)
3. Install `poetry` with `pipx` according to the [docs](https://python-poetry.org/docs/#installing-with-pipx)
4. Configure `poetry` to create the `.venv` folder in the project dir with `poetry config virtualenvs.in-project true`
	* If you have used `pyenv` do `poetry config virtualenvs.prefer-active-python true` as per the [docs](https://python-poetry.org/docs/managing-environments/)
5. Install the dependencies with `poetry install`. This creates a poetry virtual env and installs the dependencies in it.
6. Run `pre-commit install` to install the pre-commit hooks. To activate the virtual env you could use `poetry shell`, though I wouldn't recommend it in a Makefile context as it [creates a subshell](https://python-poetry.org/docs/cli/#shell). I would recommend doing `poetry run pre-commit install` as this [executes the command in the virtualenv](https://python-poetry.org/docs/cli/#run).
7. If for some reason you need to restart the shell inform the user to execute `exec $SHELL`

Isak-Bego:
#715 Adds automatic installation to the py-api

## Modifies the installation scripts and the Makefile:
Edited the installation scripts to be compliant with the new project directory and fixed some minor issues that arised from the prefix of homebrew (it has a different one for arm and x86 archs). The project should install without issues on a mac. Moreover, I edited the makefile script for installing python and fixed the paths of the other scripts. Runs on my machine :)

resolves #715

## How to verify:

- [ ] Follow the installations scripts for mac
- [ ] Switch the branch to 715-auto-install-py-api
- [ ] run the install_osx.sh script
- [ ] then run make post_osx


Makefile
```diff
@@ -1,29 +1,24 @@
...
+# 	cd ./services/web/ && npm run start
 
 .PHONY: run-prod
 run-prod:
-	cd ./packages/web/ && npm run prod
-
-.PHONY: run-api
-run-api:
-	cd ./packages/api/ && go run main.go
+	cd ./services/web/ && npm run prod
```
asynchroza, line 16: Do these commands still exist after @Aleks1908 reinitialized the project?
I recommend discarding all of those for now and adding them only as needed.

Makefile
```diff
@@ -65,42 +58,42 @@ gum:
...
-	fi \
+		echo "No supported package managers found."; \
+		exit 1; \
+	fi; \
+	pipx ensurepath; \
+	pipx install poetry; \
+	poetry config virtualenvs.in-project true; \
+	cd ./services/py_api/; \
+	poetry install; \
+	poetry run pre-commit install;
```
asynchroza, line 85: How does this work? @Aleks1908 set up husky which is a pre-commit hook manager.

Would this overwrite the pre-commit hook which is already in place?
I think I had figured out something as to account for both python and node environments in the project. You can get inspired by it. 

https://github.com/AUBGTheHUB/monolith/blob/05c73aaf1a48934f224858e0dcd8c503ff9c83aa/.husky/pre-commit#L5

Makefile
```diff
@@ -65,42 +58,42 @@ gum:
...
 
-	echo "Please, reload your shell!"
 
 .PHONY: run-py-api
 run-py-api:
-	cd packages/py-api && poetry run start
+	cd services/py_api && poetry run start
```
asynchroza, line 90: What's the reason behind the renaming? Usually top level packages especially libs are named with dashes instead of underscores as to avoid confusion between the core code which is initialized with an `init` file and the outer layer which holds all of the config files for managing the environment, etc..

install_osx.sh
```diff
@@ -1,38 +1,19 @@
 #!/bin/sh
```
asynchroza, line 1: Make a backup of the ~/.zshrc file before writing to it. Also, in order to not overwrite previous backups, name should be unique (probably just include date). 

.zshrc-backup-(day-today)

install_osx.sh
```diff
@@ -1,38 +1,19 @@
...
+	echo "\nexport NVM_DIR=\"$HOME/.nvm\"
+	[ -s \"$(brew --prefix)/opt/nvm/nvm.sh\" ] && \. \"$(brew --prefix)/opt/nvm/nvm.sh\"  # This loads nvm
+	[ -s \"$(brew --prefix)/opt/nvm/etc/bash_completion.d/nvm\" ] && \. \"$(brew --prefix)/opt/nvm/etc/bash_completion.d/nvm\"  # This loads nvm bash_completion" >> ~/.zshrc
     
-    cd $HOME/go/src
+    cd $HOME
 
 	git clone git@github.com:AUBGTheHUB/monolith.git
 
     cd monolith
```
asynchroza, line 15: Can you fix the tab formatting issues here?

asynchroza:
Changes look good for the most part but if I have to be honest I believe we should totally ditch this approach. 
Dev containers is the way to go and support only Linux envs. 

I've had one too many issues with these scripts because even when run on the same machines, if somebody has done something out of the ordinary to their shells, they fail and there's no failover. Ask @DemonHashira how many times I've had to fix his env. 

With a dev container, you would always have absolutely the same setup - no matter osx, linux or windows.

Makefile
Isak-Bego: Yeah I see. This one was supposed to set up the pre-commit hook for python.

Makefile
Isak-Bego: Me & Ivan changed the name of the directory to services.

Isak-Bego:
> Changes look good for the most part but if I have to be honest I believe we should totally ditch this approach. Dev containers is the way to go and support only Linux envs.
> 
> I've had one too many issues with these scripts because even when run on the same machines, if somebody has done something out of the ordinary to their shells, they fail and there's no failover. Ask @DemonHashira how many times I've had to fix his env.
> 
> With a dev container, you would always have absolutely the same setup - no matter osx, linux or windows.

Dear Misho, 
Thank you so much for taking the time to review this. It took me a lot of time to just research about the different systems and I feel like I wrote code which I am not 100% confident will work. I created the PR so that you guys can give me opinions on wether to continue working on this or just ditch it. I will check the dev containers, since they seem to be solving all this headache. 
Thank you again, 
Isak

asynchroza:
> > Changes look good for the most part but if I have to be honest I believe we should totally ditch this approach. Dev containers is the way to go and support only Linux envs.
> > I've had one too many issues with these scripts because even when run on the same machines, if somebody has done something out of the ordinary to their shells, they fail and there's no failover. Ask @DemonHashira how many times I've had to fix his env.
> > With a dev container, you would always have absolutely the same setup - no matter osx, linux or windows.
> 
> Dear Misho, Thank you so much for taking the time to review this. It took me a lot of time to just research about the different systems and I feel like I wrote code which I am not 100% confident will work. I created the PR so that you guys can give me opinions on wether to continue working on this or just ditch it. I will check the dev containers, since they seem to be solving all this headache. Thank you again, Isak

I will provide you with the initial setup

PR #719, opened 2024-09-23
Add support for dev container 

asynchroza:
Added support for dev containers.

Updated readme and Makefile to reflect the changes.

Dockerfile is installing the following:
- [x] Install LTS node
- [x] Installing a new version of Golang. We previously used 1.19 but since we got rid of the Go API, we can bump the version to support the latest changes of the charmbracelet/gum package.

Post command is installing all necessary dependencies to run the node dependant projects:
- [x] The package.json `prepare` script is run automatically on `npm install` invocation. Hence, hooks are installed after post command.
- [x] You will have to account for the python package installations here as well.

What's missing:
- [x] Python installation in Dockerfile
- [x] Poetry installation in Dockerfile
- [x] Poetry env initialization in post command

Solves #715


Isak-Bego (commented):
Everything looks sweet! Thank you so much Misho. I should be able to continue it from now on. We really appreciate it!

.devcontainer/Dockerfile
```diff
@@ -0,0 +1,27 @@
...
+RUN wget -O go.tar.gz https://go.dev/dl/go1.23.1.linux-amd64.tar.gz
+RUN rm -rf /usr/local/go && tar -C $HOME -xzf go.tar.gz
+RUN rm -rf go.tar.gz
+RUN echo "export PATH=$PATH:$HOME/go/bin" >> $HOME/.bashrc
+
+# Install python
+RUN wget -O python.tar.xz https://www.python.org/ftp/python/3.12.6/Python-3.12.6.tar.xz
+RUN tar -xf python.tar.xz -C $HOME 
+RUN chmod +x $HOME/Python-3.12.6/configure
+RUN /$HOME/Python-3.12.6/configure
```
asynchroza, line 23: That's one way to go about it. The other is to pull `pyenv` and install Python in a similar way how `nvm` installs the `lts` node. I think the latter will give you more flexibility but I don't know if that's needed in the first place. 

I will have @IvanObreshkov have the last saying here. My suggestion is that you just add a comment saying 

```
# You could swap this for `pyenv` to enable easily hopping between python versions
``` 

or something of the sorts.

.devcontainer/postCreateCommand.sh
```diff
@@ -0,0 +1,7 @@
+#!/usr/bin/env bash
```
asynchroza, line 1: That's a weird shebang ❓

Isn't it supposed to be `/bin/bash` ?

Also, you don't really need it as you run the script by specifying what binary to execute it. 
`bash ./script`

.devcontainer/devcontainer.json
```diff
@@ -0,0 +1,35 @@
...
+	// Use 'forwardPorts' to make a list of ports inside the container available locally.
+	"forwardPorts": [
+		3000,
+		6969,
+		8000
+	],
+	// Volume containing repo is shared after image is built, therefore we cannot run any setup steps
+	// during the image creation. This is the command which gets executed once the image is built and the 
+	// volume is mounted
+	// TODO: You will have to run all `poetry install` commands here
```
asynchroza, line 19: Remove TODO comment

Isak-Bego:
Everything seems to be working as expected. @asynchroza I just understood what you meant about the pre-hooks. I had to go through some git docs for 2 hrs, just to realise something that you told me 3 days ago. Take a last look and lmk.

.devcontainer/Dockerfile
```diff
@@ -0,0 +1,47 @@
...
+RUN echo 'export PYENV_ROOT="$HOME/.pyenv"' >> ~/.bashrc
+RUN echo 'command -v pyenv >/dev/null || export PATH="$PYENV_ROOT/bin:$PATH"' >> ~/.bashrc
+RUN echo 'eval "$(pyenv init -)"' >> ~/.bashrc
+RUN exec $SHELL
+
+RUN pyenv install 3.12.6
+RUN pyenv global 3.12.6
+
+#Install pipx and poetry
+RUN apt install -y pipx
```
asynchroza, line 45: Can't you install pip through's python's pip?

.devcontainer/Dockerfile
```diff
@@ -0,0 +1,47 @@
...
+    tk-dev \
+    libxml2-dev \
+    libxmlsec1-dev \
+    libffi-dev \
+    liblzma-dev
+RUN git clone https://github.com/pyenv/pyenv.git ~/.pyenv
+RUN echo 'export PYENV_ROOT="$HOME/.pyenv"' >> ~/.bashrc
+RUN echo 'command -v pyenv >/dev/null || export PATH="$PYENV_ROOT/bin:$PATH"' >> ~/.bashrc
+RUN echo 'eval "$(pyenv init -)"' >> ~/.bashrc
+RUN exec $SHELL
```
asynchroza, line 39: I would be a little weary of this when run in a docker container. I feel like this could impose some issues.
Does it work if you source the file instead?

.devcontainer/postCreateCommand.sh
```diff
@@ -0,0 +1,7 @@
+npm install && make install-env
+npm run prepare
+make install-gum
+poetry config virtualenvs.in-project true
+make install-web
```
asynchroza, line 5: Let's break these into groups and add some comments as to make it easier for newcomers.

asynchroza (commented):
@Isak-Bego looks good for the most part, king sir head of development. Just a couple of nits.

.devcontainer/Dockerfile
Isak-Bego: I checked the pipx documentation, and if you're running Ubuntu (which our container is) it recommends apt. For now it should not be a problem.

.devcontainer/Dockerfile
Isak-Bego: Good point. Solved in 1dad3d9

.devcontainer/postCreateCommand.sh
Isak-Bego: Solved in 1dad3d9

asynchroza (commented):
@IvanObreshkov I can't approve. You will have to give this one the approval.

Isak-Bego (approved):

PR #720, opened 2024-09-27
Added Mailing service

IvanObreshkov:
* Added mailing service with Auth middleware
* Added Dockerfile


services/email-service/src/server/middleware/auth_middleware.py
```diff
@@ -0,0 +1,37 @@
...
+                return response
+
+            if os.environ["ENV"] != "PROD":
+                if request.headers.get("X-PY-API-MACHINE-AUTHORIZATION") == "DEV":
+                    response = await call_next(request)
+                    return response
+                else:
+                    return JSONResponse(content={"message": "Unauthorized"}, status_code=401)
+            else:
+                # TODO: Implement computing digest of request and comparing it with the incoming one
```
IvanObreshkov, line 29: This may be redundant if we do not expose the service to the public, which I prefer not to do. I have put it just for the good practices.

IvanObreshkov:
Closing this as we will use https://resend.com/

PR #742, opened 2024-10-28
Hotfix/ Add team_id to the Participant model

IvanObreshkov:
I have missed to add the `team_id` to the Participant model. This small addition created a whole lot of trouble due to the Mongo ObjectID not being JSON serializable. 

Apparently, when we have fields defined with _ in a Pydantic model, the validation for those fields is skipped and they are not shown in the response. This combined with multiple errors from Pydantic due to the ObjectID not being serializable and the inability to show the `id` field in the response without the _ infront (`_id`), created the need for writing a custom serialization.

To be honest, I prefer this implementation compared to the previous one as we have better IntelliSense from the IDE when we are creating an Object of type `Participant` or `Team` (it shows the fields in the constructor, but when initializing Pydantic models we don't get that). Also, we have more customization. I prefer to use Pydantic models only for Request/Response schemas.


Isak-Bego (approved):
LTGM. Good job @IvanObreshkov

PR #752, opened 2024-10-30
Save the timestamps in the database

Isak-Bego:
#Quick fix
The methods in the serializing methods in the participant and team models did not include the timestamps when building JSON object and the mongoDB document. The base model correctly generated by default a timestamp and an objectID, however we were not using them. 

## How to verify:

- [ ] Start the backend server

- [ ] Use the create endpoint to create a new admin. 

- [ ] Check the response

### DIFF
--- a/services/py-api/src/database/model/participant_model.py
+++ b/services/py-api/src/database/model/participant_model.py
@@ -17,23 +17,23 @@ class Participant(Base):
     def dump_as_mongo_db_document(self) -> Dict[str, Any]:
         return {
             "_id": self.id,
-            "created_at": self.created_at,
-            "updated_at": self.updated_at,
             "name": self.name,
             "email": self.email,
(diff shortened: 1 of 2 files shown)


services/py-api/src/database/model/participant_model.py
```diff
@@ -17,6 +17,8 @@ class Participant(Base):
     def dump_as_mongo_db_document(self) -> Dict[str, Any]:
         return {
             "_id": self.id,
+            "created_at": self.created_at,
+            "updated_at": self.updated_at,
```
IvanObreshkov, line 21: can you move them to be at last place?

Isak-Bego: Fixed in 67ceadf

IvanObreshkov (approved):

PR #764, opened 2024-11-07
Fix some nits

IvanObreshkov:


services/py-api/src/server/routes/participants_routes.py
```diff
@@ -42,7 +42,7 @@ def _handler(service: ParticipantService = Depends(_service)) -> ParticipantHand
 
 
 @participants_router.post(
-    "", responses={201: {"model": ParticipantRegisteredInTeamResponse}, 409: {"model": ErrResponse}}
+    "", status_code=201, responses={201: {"model": ParticipantRegisteredInTeamResponse}, 409: {"model": ErrResponse}}
```
IvanObreshkov, line 45: I added this to tell the Swagger docs that this is the default status code to be returned. otherwise, it puts 200 by default. and we get possible returned response codes: 200, 201, 409, 422 and we want only 201, 409, 422

Isak-Bego: Nice!

Isak-Bego (approved):
LGTBQ+

PR #767, opened 2024-11-10
#762 Write the exception traceback in the logfile

### LINKED ISSUE #762: [BUG] Log.exception does not log the exceptions to the logfile, it logs it to the stdout
## Test case:
`ENV=DEV`
`services/py-api/src/database/repository/participants_repository.py`
image

Result:
image

When `ENV=LOCAL` it works as expected:
image

## How to achieve it:
Figure it out
also pls change the LOG of the aborting transaction to be warning when the result is an err, like this:
image
This is done as we are expected to log the exception on an upper-level (for example we are retrying the create participant operation and in the create method in the `ParticipantsRepository` we have a try-except and in the except, we do LOG.exception before returning the result.) We don't want the exception to be printed twice like this (also it says that there is no exception to be logged):
image

Isak-Bego:
#762 Writes the exception traceback in the logfile for PROD and DEV environments

### What was the issue
We were using conflicting processors for structlog, this caused the issue of the exception not being printed in the file. Moreover, I noticed a weird behaviour with the way the .env file is interpreted by load_dotenv in my machine. It seems like it was using some sort of caching system to make the reading of the env variables faster and it was not listening to new changes I was making in the .env. I literally deleted the .env file and the app was still running. To solve this issue I read a little more about load_dotenv works, which you can check on the picture below: 
Screenshot 2024-11-09 at 1 50 13 PM
Apparently, if you do not specify a value for override it defaults to False, which means that the function will first look for the value of that variable in the environment (not sure of what what they mean by environment here) and only after this in the .env file. For this reason, I change the override=True, solving the issue I was facing. 

resolves #762 

## How to verify (Please test it):

- [ ] Change part of the code in our create participant method to trigger and exception

- [ ] Switch the ENV variable to DEV or PROD

- [ ] Run the application

- [ ] Send a request to the application so that it raises an exception

- [ ] Check the logs in shared/server.log

- [ ] if "excpetion" is part of the JSON object

### DIFF
--- a/services/py-api/src/database/transaction_manager.py
+++ b/services/py-api/src/database/transaction_manager.py
@@ -141,7 +141,7 @@ class TransactionManager:
             return result
 
         except Exception as e:
-            LOG.exception("Aborting transaction due to err {}".format(e))
+            LOG.warning("Aborting transaction due to err {}".format(e))
             await session.abort_transaction()
             return Err(e)
 
--- a/services/py-api/src/database/transaction_manager.py
+++ b/services/py-api/src/database/transaction_manager.py
@@ -141,7 +141,7 @@ class TransactionManager:
             return result
 
         except Exception as e:
-            LOG.warning("Aborting transaction due to err {}".format(e))
+            LOG.exception("Aborting transaction due to err {}".format(e))
             await session.abort_transaction()
             return Err(e)
 
--- a/services/py-api/src/server/logger/logger_factory.py
+++ b/services/py-api/src/server/logger/logger_factory.py
@@ -7,7 +7,6 @@ from structlog.contextvars import merge_contextvars
 from structlog.dev import ConsoleRenderer
 from structlog.processors import (
     TimeStamper,
-    ExceptionPrettyPrinter,
     StackInfoRenderer,
     CallsiteParameterAdder,
     CallsiteParameter,
@@ -83,9 +82,8 @@ def configure_app_logger(env: str) -> None:
             add_log_level,
             merge_contextvars,
             TimeStamper(fmt="%Y-%m-%d %H:%M:%S"),
-            ExceptionPrettyPrinter(),
-            StackInfoRenderer(),
             format_exc_info,
+            StackInfoRenderer(),
             PositionalArgumentsFormatter(),
             CallsiteParameterAdder(
                 {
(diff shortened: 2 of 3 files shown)


services/py-api/src/database/transaction_manager.py
```diff
@@ -141,7 +141,7 @@ async def callback(session):
             return result
 
         except Exception as e:
-            LOG.exception("Aborting transaction due to err {}".format(e))
+            LOG.warning("Aborting transaction due to err {}".format(e))
```
IvanObreshkov, line 144: Why? Aren't we losing the stack trace in this way? On line 133 we have a check for expected error. This means that the method we are calling is expected to return an err, which we have to handle. But what if some unexpected error arises via an unhandled exception? Well, we want to catch it as early as possible in this case, before it propagates up the stack trace. As this is an unexpected Exception, this means that we have a flaw in our code, and it would be nice to see where this flaw is coming from via the stack trace. As the exception is caught, the stack trace would not be displayed to the stderr unless we explicitly want it to, and we do want the stack trace for unexpected errors so that we could understand what happened. Otherwise here is what is going to happen. Some of the methods raisees an uncaught exception and it is cough here, and we just say "WARNING the transaction was aborted due to this err (and we show the final err in the stack trace, but we are missing the caused of it, the rest of the stack trace) https://docs.python.org/3/tutorial/errors.html#exception-chaining

This method returns an Err(Exception) and in the http_handler we have a check if the Err is not of concrete type for example Err(DuplikateEmailError) we return a 500 status code. And when we dig through the logs we see information, which is not helpful for properly tracing the error. I could be mistaken, but pls double check if this is what will happen.

services/py-api/src/server/logger/logger_factory.py
```diff
@@ -83,9 +82,8 @@ def configure_app_logger(env: str) -> None:
             add_log_level,
             merge_contextvars,
             TimeStamper(fmt="%Y-%m-%d %H:%M:%S"),
-            ExceptionPrettyPrinter(),
```
IvanObreshkov, line 86: This is removed as we have the `format_exc_info`, right? At least this is what the docs say " This processor is mostly for development and testing so you can read exceptions properly formatted.  It behaves like `format_exc_info` except it removes the exception data from the event dictionary after printing it." Or the StackInfoRender takes care of all of this and the exceptions are properly printed in the server.log and the stderr (in the console)

How are the exceptions printed now to the stdout when ENV=Local and how are they printed to the server.log when ENV=DEV? I think in the `server.log` they would be on one line after the `exception` key, which would make it difficult to read. Again I could be wrong, so pls provide me with screenshots as I cannot test it right now.

IvanObreshkov (approved):
LGTM

PR #769, opened 2024-11-12
#743 Implements check 2

### LINKED ISSUE #743: Implementing Capacity Check 2
## Brief description:

We want to be able to register the new team only if there is enough space in the hackathon to accommodate it. Every year we decide on a maximum capacity for the hackathon and we use that as a constant throughout our codebase. However, we will need to count all the already created and verified teams and compare the number to the max_cap to see if we can fit another one or not. Everything seems pretty easy… We just write a query that selects all the teams which are verified from the database and count them. Surprise! The process is not as simple, since we have random participants joining (participants who do not have a team & expect us to arrange them in a team). This way, we have to allocate space (by reserving a number of teams) for these verified random participants. 

## How to achieve it:

In order to get the correct number of teams which are currently occupied we have created the formula for you!
SO USE IT!

Capacity Check 2

You want to focus on this part. However here is a link where you can access the whole schema (just in case): 
https://excalidraw.com/#json=7L64_2tnTYzpwU46FXF2V,IyMBQVvENzmAa7ULXWP5Fw

You will want to create a function inside of `participants_service.py` which is a method ParticipantService class that returns true if there is enough space to add a new team and returns false if there isn’t using the formula & the check. 

Make sure to write the database queries in the `participants_repository.py` and `teams_repository.py`. Since that is the whole point of the Handler-Service-Repository pattern. ALL THE OPERATIONS WITH THE DATABASE ARE INSIDE OF A REPOSITORY CLASS. 

Moreover the rest of the business logic the calculation of the formula and the check should be done inside of the service layer inside of the function that you just created. 

Also, write a unit test for it.

Alexandr-Ignatov:
#743 Implementing Capacity Check 2 

## In participants_service.py
New Parameters in Constructor:

hackathon_cap: Maximum capacity for the hackathon, passed in during the instantiation of ParticipantService.
team_capacity: Number of participants per team, also passed during instantiation.
These parameters are stored as instance variables so they can be used within the new _check_capacity method

New _check_capacity Method:

This async method performs the "Capacity Check 2" by:
Fetching the number of verified random participants from ParticipantsRepository using get_verified_random_participants_count.
Fetching the number of verified registered teams from TeamsRepository using get_verified_registered_teams_count.
Calculating number_ant_teams using the formula number_ant_teams = R + ceil(N / c).
Comparing number_ant_teams to hackathon_cap to check if there’s enough space for a new team.
This method returns True if there is capacity, and False otherwise.

Modification to register_admin_participant:
Before proceeding with the transaction, register_admin_participant calls _check_capacity to verify that there is enough space.
If _check_capacity returns False, it returns an Err with a custom error CapacityExceededError, indicating that the hackathon capacity has been exceeded.
If there is enough space, the method proceeds with the transaction using _create_participant_and_team_in_transaction.

New CapacityExceededError Exception:
This custom exception is returned when the capacity check fails, which can help in distinguishing this specific error during error handling.

## In participants_repository.py
Explanation:
get_verified_random_participants_count:
This method performs a query to count participants who are verified and do not have a team_id, indicating that they are random participants waiting to be assigned to a team. It returns an Ok result wrapping the count, which aligns with the existing result-handling structure in the repository.

## In team_repository.py
Explanation:
get_verified_registered_teams_count:
This method performs a  query to count teams that are verified.
It returns an Ok result wrapping the count, which aligns with the result-handling structure used in the repository.
resolves #743 

## How to verify:


services/py-api/src/service/participants_service.py
```diff
@@ -8,19 +8,22 @@
...
 from src.server.schemas.request_schemas.schemas import ParticipantRequestBody
 
 
 class ParticipantService:
     """Service layer responsible for handling the business logic when registering a participant"""
 
     def __init__(
-        self, participant_repo: ParticipantsRepository, team_repo: TeamsRepository, tx_manager: TransactionManager
+        self, participant_repo: ParticipantsRepository, team_repo: TeamsRepository, tx_manager: TransactionManager,
+        hackathon_cap: int, team_capacity: int
```
IvanObreshkov, line 20: why are these added as constructor params

services/py-api/src/service/participants_service.py
```diff
@@ -8,19 +8,22 @@
...
     def __init__(
-        self, participant_repo: ParticipantsRepository, team_repo: TeamsRepository, tx_manager: TransactionManager
+        self, participant_repo: ParticipantsRepository, team_repo: TeamsRepository, tx_manager: TransactionManager,
+        hackathon_cap: int, team_capacity: int 
     ) -> None:
         self._participant_repo = participant_repo
         self._team_repo = team_repo
         self._tx_manager = tx_manager
+        self.hackathon_cap = hackathon_cap
+        self.team_capacity = team_capacity
```
IvanObreshkov, line 26: why are these added as constructor params? I think they should be extracted as constants. @Isak-Bego can also say.

services/py-api/src/database/repository/participants_repository.py
```diff
@@ -57,3 +57,9 @@ async def create(
         except Exception as e:
             LOG.exception("Participant insertion failed due to err {}".format(e))
             return Err(e)
+    
+    async def get_verified_random_participants_count(self) -> Ok[int]:
+        """Returns the count of verified participants who are not assigned to any team."""
+        query = {"verified": True, "team_id": None}
+        count = await self._collection.count_documents(query)
```
IvanObreshkov, line 64: Can this raise and Exception? if yes, then catch it and return a Result

services/py-api/src/database/repository/teams_repository.py
```diff
@@ -57,3 +57,9 @@ async def update(
 
     async def delete(self, obj_id: str, session: Optional[AsyncIOMotorClientSession] = None) -> Result:
         raise NotImplementedError()
+    
+    async def get_verified_registered_teams_count(self) -> Ok[int]:
+        """Returns the count of verified teams."""
+        query = {"verified": True}
+        count = await self._collection.count_documents(query)
```
IvanObreshkov, line 64: Can this raise and Exception? if yes, then catch it and return a Result

services/py-api/src/service/participants_service.py
```diff
@@ -1,5 +1,5 @@
 from typing import Tuple, Optional
-
+import math
```
IvanObreshkov, line 2: from math import ceil, we don't need the whole module

services/py-api/src/service/participants_service.py
```diff
@@ -35,10 +38,29 @@ async def _create_participant_and_team_in_transaction(
...
+        verified_random_participants = await self._participant_repo.get_verified_random_participants_count()
+
+        # Fetch number of verified registered teams
+        verified_registered_teams = await self._team_repo.get_verified_registered_teams_count()
+
+        # Calculate the anticipated number of teams
+        number_ant_teams = verified_registered_teams + math.ceil(verified_random_participants / self.team_capacity)
+
+        # Check against the hackathon capacity
+        return number_ant_teams < self.hackathon_cap
```
IvanObreshkov, line 54: I think this method should be in the TeamsRepository and it should be public. As we are checking if we are able to add a Team, we are counting teams. And it should be `check_hackathon_capacity`

services/py-api/src/server/exception.py
```diff
@@ -7,3 +9,8 @@ class DuplicateEmailError(DuplicateKeyError):
 
 class DuplicateTeamNameError(DuplicateKeyError):
     pass
+
+class CapacityExceededError(HTTPException):
```
IvanObreshkov, line 13: I understand why you added this, but I am not sure we need it to be Inheriting from HTTPException, Exception is enough.

```python
class CapacityExceededError(Exception):
         pass
```

services/py-api/src/service/participants_service.py
```diff
@@ -35,10 +38,29 @@ async def _create_participant_and_team_in_transaction(
...
 
     async def register_admin_participant(
         self, input_data: ParticipantRequestBody
-    ) -> Ok[Tuple[Participant, Team]] | Err[DuplicateEmailError | DuplicateTeamNameError | Exception]:
-        # TODO: Add Capacity check 2
+    ) -> Ok[Tuple[Participant, Team]] | Err[DuplicateEmailError | DuplicateTeamNameError| CapacityExceededError | Exception]:
+        # Capacity Check 2
+        has_capacity = await self._check_capacity()
+        if not has_capacity:
+            return Err(CapacityExceededError("Hackathon capacity exceeded, cannot register new team"))
```
IvanObreshkov, line 62: return just the max capacity in the body

IvanObreshkov (commented):
Also where is the logic in the http hanlder for the new Err we introduced

services/py-api/src/database/repository/participants_repository.py
```diff
@@ -57,3 +57,9 @@ async def create(
         except Exception as e:
             LOG.exception("Participant insertion failed due to err {}".format(e))
             return Err(e)
+    
+    async def get_verified_random_participants_count(self) -> int:
+        """Returns the count of verified participants who are not assigned to any team."""
+        query = {"verified": True, "team_id": None}
+        count = await self._collection.count_documents(query)
+        return count
```
IvanObreshkov, line 65:
```python
  return await self._collection.count_documents({"email_verified": True, "team_id": None})
```
We don't have a `verified` field in the db model:
image

services/py-api/src/service/participants_service.py
```diff
@@ -35,10 +35,29 @@ async def _create_participant_and_team_in_transaction(
...
+        # Capacity Check 2
+        has_capacity = await self.check_capacity_register_admin_participant_case()
+        if not has_capacity:
+            return Err(HackathonCapacityExceededError(self._team_repo.MAX_NUMBER_OF_TEAMS_IN_HACKATHON))
+
+        # Proceed with registration if there is capacity
         result = await self._tx_manager.with_transaction(self._create_participant_and_team_in_transaction, input_data)
         return result
+
+    async def check_capacity_register_admin_participant_case(self) -> bool:
```
IvanObreshkov, line 51: `_check_capacity_register_admin_participant_case` and move it before `register_admin_participant`

IvanObreshkov (dismissed):
And Fix the tests, also write a unit test for this function. And this case in the handler

services/py-api/src/database/repository/teams_repository.py
```diff
@@ -57,3 +61,10 @@ async def update(
 
     async def delete(self, obj_id: str, session: Optional[AsyncIOMotorClientSession] = None) -> Result:
         raise NotImplementedError()
+    
+    async def get_verified_registered_teams_count(self) -> int:
+        """Returns the count of verified teams."""
+        query = {"verified": True}
+        count = await self._collection.count_documents(query)
+        return count
```
IvanObreshkov, line 69: https://github.com/AUBGTheHUB/monolith/pull/769#discussion_r1838875826

services/py-api/src/service/participants_service.py
Alexandr-Ignatov: Ivan, in your comment just below, you ask to fix the tests and write unit tests for "this" function, as well as, "And this case in the handler". 

Can you elaborate on which tests should I fix, and what does "And this case in the handler" means. Also, do I understand correctly that I need to write a unit test for "_check_capacity_register_admin_participant_case"?

IvanObreshkov: @Alexandr-Ignatov, again, as I said in the group in Messenger, there was a problem with your commits, talk with @Isak-Bego about how to proceed with this PR. I suggest we just close it, delete this branch, and start all over again, as there aren't a lot of changes, and it has become a mess. But @Isak-Bego will tell.

services/py-api/src/service/participants_service.py
Alexandr-Ignatov: Fixed f33e8f7a85dbefbcda9a089a8ba5a1d3e7e93e9b

services/py-api/src/service/participants_service.py
Alexandr-Ignatov: Fixed f33e8f7a85dbefbcda9a089a8ba5a1d3e7e93e9b

services/py-api/src/service/participants_service.py
Alexandr-Ignatov: Fixed f33e8f7a85dbefbcda9a089a8ba5a1d3e7e93e9b

services/py-api/src/database/repository/participants_repository.py
Alexandr-Ignatov: Fixed f33e8f7a85dbefbcda9a089a8ba5a1d3e7e93e9b

services/py-api/src/database/repository/teams_repository.py
Alexandr-Ignatov: Fixed f33e8f7a85dbefbcda9a089a8ba5a1d3e7e93e9b

services/py-api/src/database/repository/teams_repository.py
```diff
@@ -61,10 +60,9 @@ async def update(
 
     async def delete(self, obj_id: str, session: Optional[AsyncIOMotorClientSession] = None) -> Result:
         raise NotImplementedError()
-    
+
     async def get_verified_registered_teams_count(self) -> int:
         """Returns the count of verified teams."""
-        query = {"verified": True}
+        query = {"email_verified": True}
```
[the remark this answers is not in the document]
iobreshkov-qb: And why this thing is not  a single line as mentioned in the previous comments:
```python
 return await self._collection.count_documents(...)
```

services/py-api/src/database/repository/participants_repository.py
```diff
@@ -57,9 +57,9 @@ async def create(
         except Exception as e:
             LOG.exception("Participant insertion failed due to err {}".format(e))
             return Err(e)
-    
+
     async def get_verified_random_participants_count(self) -> int:
         """Returns the count of verified participants who are not assigned to any team."""
-        query = {"verified": True, "team_id": None}
+        query = {"email_verified": True, "team_id": None}
         count = await self._collection.count_documents(query)
```
iobreshkov-qb: Single line pls:
```python
return await self._collection.count_documents(...)
```

services/py-api/src/database/repository/participants_repository.py
```diff
@@ -57,3 +57,9 @@ async def create(
         except Exception as e:
             LOG.exception("Participant insertion failed due to err {}".format(e))
             return Err(e)
+
+    async def get_verified_random_participants_count(self) -> int:
+        """Returns the count of verified participants who are not assigned to any team."""
+        query = {"email_verified": True, "team_id": None}
+        count = await self._collection.count_documents(query)
+        return count
```
iobreshkov-qb, line 65: This should be a single line as I already mentioned:
```python
return await self._collection.count_documents(...)
```

iobreshkov-qb:
AND HAS SOMEBODY EVEN TESTED THIS??????????

iobreshkov-qb:
@Alexandr-Ignatov also fix your PR description

services/py-api/src/service/participants_service.py
Alexandr-Ignatov: Fixed ✅✅✅✅

Isak-Bego (approved):
LGTM

PR #771, opened 2024-11-13
#770 Rotation of files not working properly

### LINKED ISSUE #770: [BUG] Rotation of logfiles is not working as expected
## Brief description:
When we are in "DEV" and "PROD" environments we want to save the logs of our application in the respective deployed machine. The current version works as expected until the logfile is rotated. Once this event happens, the uvicorn logger starts logging into the new file, however the structlog logger continues logging to the old one although it has passed the size limit.

logger configuration functions are called in server_config.py
implementation of logger configuration is located in logger_factory.py
The log file for "DEV" and "PROD" environments is saved in py-api/shared/server.log

## How to achieve it:
With the current research that I have done we can either solve it using [logrotate](https://betterstack.com/community/guides/logging/how-to-manage-log-files-with-logrotate-on-ubuntu-20-04/) or we can use the class that is provided by python called [RotatingFileHandler](https://docs.python.org/3/library/logging.handlers.html#rotatingfilehandler). In the latter case, we want to create a single instance of this class that can be used by both the structlog and uvicorn loggers

Isak-Bego:
Fix #770

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -6,9 +6,12 @@
 
 .vscode/dryrun.log
 
+services/py-api/shared/*.log.*
+
 .files_for_commit.txt
 
 node_modules
 
 **/.DS_Store
+
 /.idea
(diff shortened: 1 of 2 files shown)


IvanObreshkov (approved):
Excellent work

PR #772, opened 2024-11-13
#744 Added mail service for resend

### LINKED ISSUE #744: feature: add resend for sending emails
We have decided to use [Resend](https://resend.com) as our email service. What we need to create is a Facade Pattern over it so that we are not tied directly to its API, and we could easily swap it in the future if needed. 

The interface that we need to expose for callers should accept: 
- receiver (the email of the person receiving the email)
- Subject (str)
- Body of the email. The tricky part here is that we need to support both simple text and HTML email templates.

You can find the email templates from last year [here](https://github.com/AUBGTheHUB/monolith/tree/master/packages%2Fpy-api%2Fresources%2Femail_templates) to get more context

[Here](https://github.com/AUBGTheHUB/monolith/blob/05c73aaf1a48934f224858e0dcd8c503ff9c83aa/packages/py-api/py_api/functionality/hackathon/mailing_functionality.py) is the implementation from last year. I am putting it just for context again, as we will write a better one.

cl3vy:
#744 Added `MailService` Interface and `ResendMailService` implementation (#744)

This PR introduces a new `ResendMailService` implementation of the `MailService` interface, enabling email functionality for participant confirmations and event communications using Resend. This includes both generic email sending and tailored participant-specific emails for successful confirmations and pending confirmations, with support for HTML templates to enhance the email presentation.

resolves #744  

### Key Features

1. **`send_email`**: A general-purpose email method that sends to a specified recipient with options for both text and HTML bodies.
2. **`send_participant_successful_confirmation_email`**: Sends a confirmation email to participants after successful acceptance into the event, including participant details, team name, and an invite link.
3. **`send_participant_confirmation_email`**: Notifies participants of required action to confirm their application, including a confirmation link and participant details.
4. **HTML Template Integration**: Supports HTML email templates with placeholders.

### Notes

- The email service uses Resend as the backend, with the API key pulled from the `.env`.

## How to verify:

- [ ] Will add these later.


services/py-api/src/service/mail_service/resend_service.py
```diff
@@ -0,0 +1,95 @@
...
+            "subject": subject,
+        }
+
+        if body_html:
+            params["html"] = body_html
+        elif body_text:
+            params["text"] = body_text
+
+        try:
+            email = self.client.Emails.send(params)
```
IvanObreshkov, line 44: Running a sync method inside an async ctx will block the event loop, and we will not be able to process any requests until the task is completed. The tasks that run in an async ctx should be able to yield control back to the event loop. How to avoid this situation:

1. Preferred approach: remove the async and write in the docs that this method is blocking and should be executed via [BackgroundTasks.add_task()](https://fastapi.tiangolo.com/tutorial/background-tasks/)

2. BackgroundTasks uses [to_thread](https://docs.python.org/3/library/asyncio-task.html#asyncio.to_thread) under the hood so you can write the wrapper yourself to send the email asynchronously. 

3. If, for some reason, we need to have a cancelation mechanism due to some deadline being exceeded (timeout) via [wait_for](https://docs.python.org/3/library/asyncio-task.html#asyncio.wait_for), the first and second approaches will not be useful. The wait_for will cancel the future returned (not the thread) but unfortunately the thread will continue running indefinitely, which could create strange behavior. You can check more [here](https://github.com/python/cpython/issues/87185). Instead we should be using [asyncio.loop.run_in_executor](https://docs.python.org/3.10/library/asyncio-eventloop.html#asyncio.loop.run_in_executor) to schedule the background task and get a future we can use `wait_for` to wait on the event-loop.

Something else, I am not sure if `send` has a retry logic underneath, but as we are working with remote systems is important to have a retry logic on TransientErrors or 500 response codes. You can grab some inspiration from here: https://github.com/AUBGTheHUB/monolith/blob/30fe96921c1b8f212274389057d5bc39b5aed654/services/py-api/src/database/transaction_manager.py#L43

services/py-api/src/service/mail_service/resend_service.py
```diff
@@ -0,0 +1,95 @@
...
+            # TODO: I need to actually log here
+            print("Email sent successfully: ", email)
```
IvanObreshkov, line 46: Then, use a logger, preferably LOG.info https://github.com/AUBGTheHUB/monolith/blob/30fe96921c1b8f212274389057d5bc39b5aed654/services/py-api/src/database/repository/participants_repository.py#L15

services/py-api/src/service/mail_service/resend_service.py
```diff
@@ -0,0 +1,95 @@
...
+        except Exception as e:
+            raise RuntimeError(f"Failed to send email: {str(e)}") from e
```
IvanObreshkov, line 48: We use the [Result pattern](https://www.milanjovanovic.tech/blog/functional-error-handling-in-dotnet-with-the-result-pattern) via [this](https://github.com/rustedpy/result) across our code base, so I suggest you to do: 
`return Err(...)` and log the exception with `LOG.exception`. Here is a good example:
https://github.com/AUBGTheHUB/monolith/blob/30fe96921c1b8f212274389057d5bc39b5aed654/services/py-api/src/database/transaction_manager.py#L143-L146

otherwise, as this method with your code version raises exceptions, you should declare this in the docs so that the caller knows that it has to do some try except action. But the caller should read the docs :D, using `Result` forces you to think more about error handling, as you know by the method definition that it could return an `Err`.  As mentioned in the article "Exceptions are for exceptional situations."

services/py-api/src/service/mail_service/resend_service.py
```diff
@@ -0,0 +1,95 @@
...
+    ) -> None:
+        try:
+            body_html = load_email_participant_html_template(participant, team_name, invite_link)
+        except ValueError as e:
+            raise ValueError(f"Error loading the HTML template: {str(e)}") from e
+
+        try:
+            await self.send_email(receiver=receiver, subject=subject, body_html=body_html, body_text=body_text)
+        except Exception as e:
+            raise RuntimeError(f"Failed to send email: {str(e)}") from e
```
IvanObreshkov, line 68: Use result pattern

services/py-api/src/service/mail_service/resend_service.py
```diff
@@ -0,0 +1,95 @@
...
+
+        try:
+            body_html = load_email_verify_participant_html_template(participant, team_name, confirmation_link)
+        except ValueError as e:
+            raise ValueError(f"Failed to load email template: {str(e)}")
+
+        try:
+            await self.send_email(receiver=receiver, subject=subject, body_html=body_html, body_text=body_text)
+        except Exception as e:
+            raise RuntimeError(f"Failed to send participant confirmation email: {str(e)}") from e
```
IvanObreshkov, line 95: I think you already know what I am going to say, if not look the comments on the other files.

services/py-api/src/service/mail_service/utils.py
```diff
@@ -0,0 +1,42 @@
+def load_email_participant_html_template(participant_name, team_name, invite_link):
```
IvanObreshkov, line 1:
```python
def load_email_participant_html_template(participant_name, team_name, invite_link) -> Result[str,ValueError):
```
or some other type of error. I think raising ValueErrors on missing required params is obsolete.

services/py-api/src/service/mail_service/utils.py
```diff
@@ -0,0 +1,42 @@
...
+        inviteLink=invite_link,
+        title_team=title_team,
+        title_no_team=title_no_team,
+        inviteLinkVisibility=invite_link_visibility,
+    )
+
+
+def load_email_verify_participant_html_template(participant_name, team_name, confirmation_link):
+    if not participant_name or not confirmation_link:
+        raise ValueError("Both participant_name and verify_link must be provided.")
```
IvanObreshkov, line 26: again here

IvanObreshkov:
Resolve the changes, the tests will be in another PR

services/py-api/src/service/mail_service/resend_service.py
Isak-Bego, line 44: The call is not awaited so I don't think it will work like this.
```python
email = await self.client.Emails.send(params)
```

Isak-Bego: Why would it block the event loop. I think he will only be left with email having an undefined value since it did not wait for the request to finish. Then if we are calling this async 'send_mail' method from a syncronous context we need to use the BackgroundTasks, otherwise we are blocking the flow of the application @IvanObreshkov

services/py-api/src/service/mail_service/resend_service.py
Isak-Bego: What a exceptional explanation by @IvanObreshkov! ⭐️

Isak-Bego (changes requested):
Overall good job!  As Ivan has mentioned a couple of fixes regarding the error handling using the Result pattern. We also need to address the async method send_mail and make sure that it is not blocking the flow of the program. If you can give it a quick test of the whole workflow and see if it works.

services/py-api/src/service/mail_service/resend_service.py
IvanObreshkov: Yes @Isak-Bego, @cl3vy's version of the code right now does NOT use `BackgroundTasks` and is calling the sync method `resend.Emails.send` inside a function that is declared to be async. If the caller calls this function  as it is right now and awaits it, the whole event loop will be blocked.

IvanObreshkov: Answering this:

image


WELL `resend.Emails.send(params)` IS NOT ASYNC, this why it is not awaited :D
image

cl3vy:
> Overall good job! As Ivan has mentioned a couple of fixes regarding the error handling using the Result pattern. We also need to address the async method send_mail and make sure that it is not blocking the flow of the program. If you can give it a quick test of the whole workflow and see if it works.

services/py-api/src/service/mail_service/resend_service.py
cl3vy: > Also, I guess you have checked the API response codes. Currently your code does not handle this case: https://resend.com/docs/api-reference/introduction#rate-limit

One good thing the resend SDK does is that it takes care of this.

services/py-api/src/service/mail_service/mail_service.py
```diff
@@ -0,0 +1,63 @@
...
+        Supports both text and HTML bodies.
+        :param receiver: Recipient's email address
+        :param subject: Email subject line
+        :param body_content: Body of the email, could be html or just plain text, must be specified in content_type
+        :param content_type: Defines the type of the email body, could be "html" (for html body) or "text" (for a plain text body)
+        """
+        raise NotImplementedError()
+
+    @abstractmethod
+    async def send_participant_successful_registration_email(
```
asynchroza, line 21: What does this have to do with the mailing service? 

You should have a concrete method which is handling the send operation and then you can create multiple different mailing services by extending the base concrete class which would handle the different business domains such as HackathonMailService which will be implementing the `send_participant_successful_registration_email`

asynchroza: It's perfectly fine to have an interface/abstract class for the MailService but you generally don't need it as we don't have expect to have different mailing clients - only resend.

asynchroza: MailService to be a concrete class with a concrete send method. 

HackathonMailService to extend MailService and implement all domain specific methods. No abstractions is needed whatsoever.

Isak-Bego: > It's perfectly fine to have an interface/abstract class for the MailService but you generally don't need it as we don't expect to have different mailing clients - only resend.

We want to have an abstract class so that we have a well defined interface for sending email which is independent of the mailing service we are using. It is true that for this year we are going to use Resend, however thinking about the future we want to be able to swap the mailing client without causing other necessary changes in our code base. 
As I believe in your understanding of this matter, I think you meant that we are over complicating it. Which to be honest for someone who has taken a simple course in object oriented programming should not be a problem to understand

Isak-Bego: > MailService to be a concrete class with a concrete send method.
> 
> HackathonMailService to extend MailService and implement all domain specific methods. No abstraction is needed whatsoever.

@asynchroza I agree with you on this part. I suggest the following. We can have MailService as an interface and then HacathonMailService implements this interface and adds domain specific methods. 

image

cl3vy: Well it does not make any sense because the only kind of MailService we will have is a HackathonMailService, the project specifies the domain. This is being overthinked too much, MailService of a hackathon project will specify its projects domain specific methods because we werent going to build a general MailService anyway.

Isak-Bego: I agree with you on the fact that this is the Hackathon API so we should only be sending Hackathon emails. However, it would be a more flexible design if we have a Mail Service class (instead of the interface that I have suggested above), that implements the send_mail() method with a specific mailing client (in this case Resend) and than having more specific classes extending it providing more convenient interfaces (that do not require so many parameters, like the send_participant_successful_email(), ... etc.) for specific usages as you have done. 
This would be super easy to implement now. You just separate the send_mail() form the rest of the methods on a class called Mail Service and have HackathonMailService class extend it and provide the more specific methods that use send_mail() under the hood. @cl3vy let me know what you think.

asynchroza: > It is true that for this year we are going to use Resend, however thinking about the future we want to be able to swap the mailing client without causing other necessary changes in our code base.

@Isak-Bego the concrete base class in this case acts as a facade. It doesn't make sense to do the following:

```
MailService (send_email()) (interface) -> ResendMailService (send_email()) -> HackathonMailService (methods use send_email() under the hood)
```

Instead, have the base class as the concrete implementation and a hackathon service with the domain specific knowledge which utilizes the `send_email` method of the base class.

asynchroza:
```
MailService (send_email()) (interface) -> ResendMailService (send_email()) -> HackathonMailService (methods use send_email() under the hood)
```

This would make sense if we were to implement either one of the factory methods but as you don't expect to use more than one email client at the same time, this just overcomplicates the code unnecessarily.

asynchroza: Shameless plug - read this [about the facade pattern](https://asynchroza.github.io/posts/facade-pattern/)

cl3vy: @asynchroza MailService is an interface because we might change the mail client in the future, not because there might be other kinds of MailServices other than a HackathonMailService, considering this is a hackathon project the only kind of MailService we will have is a HackathonMailService. send_participant_successful_registration_email method makes sense to be in the MailService because MailService in itself is the HackathonMailService, it's just not named that way because it's implicit considering that we don't offer any other kinds of mail services.

asynchroza: Currently, the project's sole focus is on the "hackathon" domain. However, this doesn't mean the team lacks plans for future expansion. In fact, two years ago, we had an admin panel specifically designed for managing events and hubbers.

From a clean code perspective, if we were to strictly adhere to all guidelines, `MailClient` would ideally be an interface implemented by specific clients such as `GMail`, `Outlook`, and `Resend`. The `HackathonMailingService` would then follow the dependency injection pattern by accepting an instance of MailClient.

```js
const mailClient = ResendMailClient();
const hackathonMailingService = HackathonMailingService(mailClient);
```

My suggestion to define `MailClient` as a concrete implementation, mimicing the facade pattern, was aimed at reducing your workload. I find it unlikely that The Hub will be using multiple mailing clients simultaneously. Therefore, having a single concrete MailClient with a `send_mail` method (protected if base is inherited, public if used in DI), shared across all MailService classes, represents the most efficient and extensible solution for the effort invested.

```js
HackathonMailingService extends MailClient {
   func send_verification_email() {
       this->send_emai(...)
   }
}
```

or

```js
HackathonMailingService {
   HackathonMailingService(emailClient) {
      this->emailClient = emailClient;
   }
   
   func send_verification_email() {
       this->send_emai()
   }
}

var emailClient := MailClient();
var hackathonMailingService = HackathonMailingService(emailClient);
```

asynchroza: > And when we say ResendMailService we mean ResendHackathonMailService if we want to look at it in that way.

This way of structure your classes already makes it too convoluted and in turn making it hard to extend or replace in the future.

asynchroza: Generally, it's up to you how you approach the problem. If it works, it works.

cl3vy: Makes sense, will take a look at implementing what you said.

cl3vy (commented):
I need a second eye on this commit @Isak-Bego , @IvanObreshkov , @asynchroza

services/py-api/src/service/mail_service/resend_service.py
```diff
@@ -19,35 +23,67 @@
...
+        if result.is_err():
+            LOG.error(f"Participant with email {receiver} not found.")
+            return Err(ParticipantNotFoundError())
+
+        participant = result.ok_value
+        # Check rate limit
+        if participant.last_sent_email:
+            if datetime.now() - participant.last_sent_email < timedelta(seconds=RATE_LIMIT_SECONDS):
+                LOG.error(f"Rate limit exceeded for participant {participant.id} with email {participant.email}.")
+                return Err(EmailRateLimitExceededError())
```
Isak-Bego, line 57: Let's create a new route for the purpose of sending the emails. We can extract this logic out of the MailingService, since in the end of the day it is not the responsibility of the mailing service to check if the participant exist or update it. 

- This way we make our code more easily testable. 
- We separate the concerns. 
- We give front-end an endpoint they can trigger to enable the resend logic.

services/py-api/src/service/mail_service/resend_service.py
```diff
@@ -19,35 +23,67 @@
...
+        retry_delay = INITIAL_RETRY_DELAY_SECONDS
+        for attempt in range(1, MAX_RETRIES + 1):
+            try:
+                # Send the email asynchronously
+                email: resend.Email = await to_thread(self.client.Emails.send, params)
+                LOG.info(f"Email sent successfully: {email}")
+
+                # Update participant's last sent email timestamp
+                participant.last_sent_email = datetime.now()
+                await self.participant_repository.update(obj_id=participant.id, updated_data=participant)
```
Isak-Bego, line 77: For the same reason as above, let's do it separately in a new PR.

services/py-api/src/service/mail_service/resend_service.py
```diff
@@ -19,35 +23,67 @@
...
+
+                return Ok(email)
+            except Exception:
+                LOG.warning(f"Attempt {attempt} failed. Retrying in {retry_delay} seconds...")
+                await sleep(retry_delay)
+                retry_delay *= 2  # Exponential backoff
+
```
Isak-Bego, line 84: Are all the Errors that resend raises retryable. If you can't find any information regarding that I guess let's keep it like this.

services/py-api/src/service/mail_service/resend_service.py
```diff
@@ -66,20 +102,16 @@ async def send_participant_successful_registration_email(
...
-            receiver: str,
-            subject: str,
-            participant: str,
-            team_name: str,
-            confirmation_link: str,
-    ) -> Result[resend.Email, str]:
+        self,
+        receiver: str,
+        subject: str,
+        participant: str,
```
Isak-Bego, line 111: Lets change this to participant_name or something similiar, because currently is too general.

services/py-api/src/service/mail_service/resend_service.py
```diff
@@ -19,35 +23,67 @@
...
         self.client = resend
+        self.participant_repository = participant_repository
 
     async def send_email(
         self, receiver: str, subject: str, body_content: str, content_type: Literal["text", "html"] = "text"
-    ) -> Result[resend.Email, str]:
+    ) -> Result[resend.Email, ParticipantNotFoundError | EmailRateLimitExceededError | str]:
+        # Check if participant exists
+        result = await self.participant_repository.fetch_by_email(receiver)
+        if result.is_err():
```
Isak-Bego, line 48: Let's change it to: 

```Python
if is_err(result):
```

for consistency

services/py-api/src/service/mail_service/resend_service.py
```diff
@@ -66,20 +102,16 @@ async def send_participant_successful_registration_email(
...
     async def send_participant_verification_email(
-            self,
-            receiver: str,
-            subject: str,
-            participant: str,
-            team_name: str,
-            confirmation_link: str,
-    ) -> Result[resend.Email, str]:
+        self,
+        receiver: str,
```
Isak-Bego, line 109: Let's find a better suited name for this argument as well. It is too general and makes you wonder what you should actually put there.

Isak-Bego (approved):
Looks Good To Me!

PR #783, opened 2024-11-16
#745 Encode And Decode Methods for JWT 

### LINKED ISSUE #745: Add JWT Encoding and Decoding
## Brief description:
We need a way to encode information into a JWT token and have a way to decode tokens. You can learn more about JWT tokens [here](https://jwt.io/introduction)
## How to achieve it:
Use ONLY the [pyjwt lib](https://pyjwt.readthedocs.io/en/stable/), we don't want unnecessary dependencies

The `encode` method should accept a payload, preferably a TypedDict, and encode it using the signing secret. It should return the encoded JWT token as a str. If the encode method could throw errors, and catch them and if this is the case the return type should be `Result[OK|Err]`

You can generate a secret key using this, but please **MAKE SURE YOU DON'T LEAK IT**. Store it in an env variable:
```bash
python -c 'import secrets; print(secrets.token_hex())'
```

The `decode` method should return the decoded JWT token, preferably as a TypedDict. It should catch `ExpiredSignatureError`, `InvalidSignatureError`, and  `DecodeError`. Check what errors could the decode method throw and catch them. The return type of the function should be `Result[OK|Err]`

plamenapavlova:
#745 Added encoding And decoding Methods for JWT 

## Added encode and decode functions:
Encode and decode functions with a check for decode errors. A class TypedDict including the participant data.

resolves #745

## How to verify:

### DIFF
--- a/services/py-api/src/utils.py
+++ b/services/py-api/src/utils.py
@@ -1,7 +1,9 @@
+from os import environ
 from threading import Lock
-from typing import Any, Dict
-
+from typing import Any, Dict, TypedDict
+from jwt import DecodeError, ExpiredSignatureError, InvalidSignatureError, decode, encode
 import httpx
+import jwt
 from structlog.stdlib import get_logger
 
 LOG = get_logger()
@@ -43,3 +45,37 @@ class AsyncClient(metaclass=SingletonMeta):
             raise RuntimeError("The AsyncClient has not been initialized!")
 
         await self._async_client.aclose()
+
+class JwtUtility:
+    class UserData(TypedDict):
+        id: int
+        name: str
+        email: str
+        is_admin: bool
+        email_verified: bool
+        team_id: int
+        created_at: str
+        updated_at: str
+         
+
+    @staticmethod
+    def encode(data: UserData) -> str:
+            return encode(data,key=environ["SECRET_KEY"])
+    
+    @staticmethod
+    def decode(token: str) -> Dict[str, Any]:
+     try:
+        decoded_payload=jwt.decode(token, SECRET_KEY, algorithms=["HS256"])
+        return decoded_payload
+     except ExpiredSignatureError:
+            print("The token has expired.")
+            return {"error": "TokenExpired"}
+     except InvalidSignatureError:
+            print("The token has an invalid signature.")
+            return {"error": "InvalidSignature"}
+     except DecodeError:
+            print("There was an error decoding the token.")
+            return {"error": "DecodeError"}
+    
+
+print(JwtUtility.encode({"str":"fdsafsad"}))


services/py-api/src/utils.py
```diff
@@ -1,7 +1,9 @@
+from os import environ
 from threading import Lock
-from typing import Any, Dict
-
+from typing import Any, Dict, TypedDict
+from jwt import DecodeError, ExpiredSignatureError, InvalidSignatureError, decode, encode
 import httpx
+import jwt
```
Isak-Bego, line 6: There is no reason to import the whole jwt library again. Add whatever dependency you need from jwt in the import statement in line no.4

services/py-api/src/utils.py
```diff
@@ -43,3 +45,37 @@ async def stop(self) -> None:
...
+class JwtUtility:
+    class UserData(TypedDict):
+        id: int
+        name: str
+        email: str
+        is_admin: bool
+        email_verified: bool
+        team_id: int
+        created_at: str
+        updated_at: str
```
Isak-Bego, line 58: Why is there class nesting? This is not right. We already have a model for both the Participant and Team, including schemas for requests and responses. So you don't need to create your own model. Navigate to 

```Python
cd /services/py-api/database/model/participant_model.py
```

The participant adds the participant specific properties on top of the base model. So you should have all the properties in UserData available in the participant_model. Just import it. 

```python
from src.database.model.participant_model import Participant
```

services/py-api/src/utils.py
```diff
@@ -43,3 +45,37 @@ async def stop(self) -> None:
...
+            return {"error": "TokenExpired"}
+     except InvalidSignatureError:
+            print("The token has an invalid signature.")
+            return {"error": "InvalidSignature"}
+     except DecodeError:
+            print("There was an error decoding the token.")
+            return {"error": "DecodeError"}
+    
+
+print(JwtUtility.encode({"str":"fdsafsad"}))
```
Isak-Bego, line 81: I see that you are trying to test it here. but it won't really work if you just start the backend server. Since the backend server is and HTTP server it will only trigger methods based on the requests in gets in its endpoints. Meaning, you have to go to py-api/server/routes/utility_routes.py and add a route that triggers this function 

```Python
from src.utils.py  import JwtUtility
@utility_router.get("/testjwt", responses={200: {"model": PongResponse}, 503: {"model": ErrResponse}})
async def testJWT() -> Any:
    return await JwtUtility.encode({"str":"fdsafsad"})
```

Open postman and do a GET request in https://localhost:8080/testjwt and it should trigger your function and you will be able to see the message in your console.

services/py-api/src/utils.py
```diff
@@ -43,3 +45,37 @@ async def stop(self) -> None:
...
+        is_admin: bool
+        email_verified: bool
+        team_id: int
+        created_at: str
+        updated_at: str
+         
+
+    @staticmethod
+    def encode(data: UserData) -> str:
+            return encode(data,key=environ["SECRET_KEY"])
```
Isak-Bego, line 63: Both of the methods the one from the jwt library and the one that you have defined on your own have the same name. This is ambiguous! please change the name of your own method to encode_data. Also, use the Participant form the participant model as a type for the data, as I explained in the previous comment.

services/py-api/src/utils.py
```diff
@@ -43,3 +45,37 @@ async def stop(self) -> None:
...
+    
+    @staticmethod
+    def decode(token: str) -> Dict[str, Any]:
+     try:
+        decoded_payload=jwt.decode(token, SECRET_KEY, algorithms=["HS256"])
```
Isak-Bego, line 68: Same thing with the naming here. change the method that you define yourself to decode_data

Isak-Bego: Put the algorithm in the .env variable and get it from there so that is safer. We are not showing all our visitors of the public repository the type of algorithm we are using for the Jwt secret

services/py-api/src/utils.py
```diff
@@ -43,3 +45,37 @@ async def stop(self) -> None:
...
+        return decoded_payload
+     except ExpiredSignatureError:
+            print("The token has expired.")
```
Isak-Bego, line 71: You can use the structlog to log this message. Similar to what is done in : 
```Python
except ConnectionFailure as cf:
            LOG.exception("Pinging db failed due to err {}".format(cf))
            return Err("Database is unavailable!")
```

This time you log the message as a warning instead of exception and something like: 

```Python
LOG.warning("Failed to decode the jwt token due to blah blah")
return Err("The jwt token has expired.")
```

Do the same thing for all the following exceptions.

services/py-api/src/utils.py
iobreshkov-qb: No need for this, I suggest @plamenapavlova write "integration-test" for this as writing a unit test is not going to do much work for us. 
What the intergration test should do:

1. Encode some info and then try to decode it  -> it should return success
2. Encode invalid info and check if returns an error
3. Decode an expired jwt token (you could have it hardcoded in a const)
4. Decode a jwt token which is signed with a different secret (you could have it the test-secret in a const)
5. Decode a malformed JWT

services/py-api/src/utils.py
```diff
@@ -43,3 +45,37 @@ async def stop(self) -> None:
...
+        created_at: str
+        updated_at: str
+         
+
+    @staticmethod
+    def encode(data: UserData) -> str:
+            return encode(data,key=environ["SECRET_KEY"])
+    
+    @staticmethod
+    def decode(token: str) -> Dict[str, Any]:
```
Isak-Bego, line 66: Change the return type to Result[OK, Err] similar to: 
```Python
    self, input_data: ParticipantRequestBody
    ) -> Result[Tuple[Participant, Team], DuplicateEmailError | DuplicateTeamNameError | Exception]:
        # TODO: Add Capacity check 2
        result = await self._tx_manager.with_transaction(self._create_participant_and_team_in_transaction, input_data)
        return result
```

services/py-api/src/utils.py
plamenapavlova: Should I import the Secret key and the Algorithm in the beginning in some way, as there are not defined and Pylance doesn't recognise them?

services/py-api/src/utils.py
plamenapavlova: In the decode method I use "jwt.decode" and without importing the whole library it writes it is not defined

Isak-Bego: Also please install the dependecy using poetry. You have not added jwt in the dependencies. 

```
poetry add pyjwt
```

Isak-Bego: do from jwt import jwt instead of import jwt

services/py-api/src/utils.py
Isak-Bego: Yeah absolutely, I think you can just go ahead and do that. Thank you Plamena! Also about the testing part, as Ivan suggested just write an integration test as he describes it!

plamenapavlova (commented):
Awaiting review

services/py-api/src/utils.py
asynchroza: > There is no reason to import the whole jwt library again. Add whatever dependency you need from jwt in the import statement in line no.4

To clarify, as this comment might seem a bit confusing: Both from `<package> import ...` and `import <package>` load the entire package into memory. However, there's a key difference in how you access the package's attributes.

With `import <package>`, you can access any attribute of the package by explicitly using <package>.<attribute>.

With `from <package> import ...`, the imported attributes are added directly to the current namespace. This allows you to access them without prefixing them with the package name. However, this approach can clutter the namespace, potentially leading to conflicts or ambiguity if multiple attributes with the same name are imported from different sources.

Having experienced both sides of this issue, I generally advise against using `from ... import ...`

services/py-api/src/utils.py
```diff
@@ -43,3 +49,26 @@ async def stop(self) -> None:
             raise RuntimeError("The AsyncClient has not been initialized!")
 
         await self._async_client.aclose()
+
+class JwtUtility:
+ 
+    @staticmethod
+    def encode_data(data: Participant) -> str:
```
IvanObreshkov, line 56:
```python
def encode_data(data: JwtUserData) -> str:
```
We don't want to encode the whole Participant obj into a JWT token, as it could have sensitive information, such as their email.

services/py-api/src/utils.py
```diff
@@ -43,3 +49,26 @@ async def stop(self) -> None:
...
+            return Err("The jwt token has expired.")
+     except InvalidSignatureError:
+            LOG.warning("Failed to decode the JWT token: The token has an invalid signature.")
+            return Err("The jwt token is invalid.")
+     except DecodeError:
+            LOG.warning("Failed to decode the JWT token: There was an error decoding the token.")
+            return Err("There was an error decoding the token.")
+    
+a= JwtUtility.decode_data("decoding data...")
+a.ok_value
```
IvanObreshkov, line 74: Remove this.

services/py-api/src/server/routes/utility_routes.py
```diff
@@ -17,3 +19,8 @@ async def ping(
     response: Response, handler: UtilityHandlers = Depends(create_utility_handler)
 ) -> PongResponse | ErrResponse:
     return await handler.ping_services(response)
+
+#Test encode function
+@utility_router.get("/testjwt", responses={200: {"model": PongResponse}, 503: {"model": ErrResponse}})
+async def testJWT() -> Any:
+    return await JwtUtility.encode({"str":"fdsafsad"})
```
IvanObreshkov, line 26: Remove this, we don't need a separate route to test the jwt functionality.

I suggest @plamenapavlova write an "integration test" for this, as writing a unit test is not going to do much work for us. (we want to test the real behavior of the jwt utils you have written)

What the integration test should do:

Encode some info and then try to decode it -> it should return success
Encode invalid info and check if returns an error
Decode an expired jwt token (you could have it hardcoded in a const)
Decode a jwt token which is signed with a different secret (you could have it the test-secret in a const)
Decode a malformed JWT

IvanObreshkov:
@plamenapavlova pls add your updated `pyproject.toml` and `poetry.lock` files in order for the tests to pass. Also, Add the env var `SECRET_KEY` to the `env.sample` and how to generate it as a comment. You have the code for the generation inside your ticket.

services/py-api/src/utils.py
Isak-Bego: @asynchroza Thank you for the correction!

plamenapavlova: Fixed in be223527926be992bf6729e80ee043728b8fa1c5

services/py-api/src/utils.py
plamenapavlova: Fixed in be223527926be992bf6729e80ee043728b8fa1c5

services/py-api/src/utils.py
plamenapavlova: Fixed in be223527926be992bf6729e80ee043728b8fa1c5

services/py-api/src/utils.py
plamenapavlova: Fixed in be223527926be992bf6729e80ee043728b8fa1c5

services/py-api/src/utils.py
plamenapavlova: Fixed in be223527926be992bf6729e80ee043728b8fa1c5

services/py-api/src/utils.py
plamenapavlova: Fixed in be223527926be992bf6729e80ee043728b8fa1c5

services/py-api/src/utils.py
plamenapavlova: Fixed in be223527926be992bf6729e80ee043728b8fa1c5

IvanObreshkov:
> > @plamenapavlova pls add your updated `pyproject.toml` and `poetry.lock` files in order for the tests to pass. Also, Add the env var `SECRET_KEY` to the `env.sample` and how to generate it as a comment. You have the code for the generation inside your ticket. And Update your branch.
> 
> The SECRET_KEY variable already exists in the .env.sample. Moreover, I added jwt in the poetry config

Nice one. But I believe you were looking at the wrong .env.sample (in monolith). The right one is in https://github.com/AUBGTheHUB/monolith/blob/main/services/py-api/.env.sample

services/py-api/src/utils.py
```diff
@@ -1,8 +1,10 @@
 from threading import Lock
```
asynchroza, line 1: Hey, hey @plamenapavlova @IvanObreshkov @Isak-Bego . This class seems to be really coupled with the hackathon business logic. What if you decide to use it for an admin panel where hubbers should use jwt to authenticate and authorize themselves? Will their tokens use the same schema?

```
            if "sub" in decoded_token and "is_admin" in decoded_token and "team_id" in decoded_token:
                return Ok(cast(JwtUserData, decoded_token))

            LOG.warning(
                "Decoded token is missing some or all of the required fields: `sub`, `is_admin`, `team_id`",
                decoded_token=decoded_token,
            )
```

services/py-api/src/utils.py
```diff
@@ -43,3 +45,40 @@ async def stop(self) -> None:
...
         await self._async_client.aclose()
+
+
+class JwtUtility:
+    """A class providing methods for encoding data in with a predefined format into a JWT token, or decoding a JWT
+    token into a predefined format (TypedDict)"""
+
+    @staticmethod
+    def encode_data(data: JwtUserData) -> str:
+        return str(encode(data, key="secret", algorithm="HS256"))
```
asynchroza, line 56: You're not using the actual secret as a key

services/py-api/src/utils.py
```diff
@@ -43,3 +45,40 @@ async def stop(self) -> None:
...
+
+    @staticmethod
+    def decode_data(token: str) -> Result[JwtUserData, str]:
```
asynchroza, line 59: Is this a common pattern within the service? Returning a string instead of an error? I can see how this is more efficient but I think it would be really nice if you establish a proper convention as to what type should always be returned. In most cases, such as in Rust, that would be an Exception. I'm pretty sure I've seen exception being returned instead of strings.

services/py-api/src/utils.py
```diff
@@ -43,3 +45,40 @@ async def stop(self) -> None:
...
+        try:
+            decoded_token: dict[str, Any] = decode(token, "secret", algorithms=["HS256"])
+
+            if "sub" in decoded_token and "is_admin" in decoded_token and "team_id" in decoded_token:
+                return Ok(cast(JwtUserData, decoded_token))
+
+            LOG.warning(
```
asynchroza, line 66: I would avoid logging all of these. They do not really provide any value in deployed environments as it's expected that these are user errors.

In the case they're not user facing errors, use `LOG.debug`

services/py-api/src/utils.py
iobreshkov-qb: > Also, a sanity check, why do you have to cast to a string?

mypy

services/py-api/src/utils.py
iobreshkov-qb: yeah, usually we return Err(Exception(...)), but I thought we could use the err messages directly in order to get displayed in the fronted.

services/py-api/src/utils.py
iobreshkov-qb: Are they not user-facing errors? For example, showing that the JWT has expired is a thing you want to show your user. What is your definition of user-facing errors?

services/py-api/src/utils.py
iobreshkov-qb: The check is here only due to the schema. You are right that we have to design it to be more generic, but we have to come up with a way to return TypedDIcts or better create custom types for the different use cases. And just pass them as Generics or something like that.

services/py-api/src/utils.py
asynchroza: Reread my comment

services/py-api/src/server/routes/utility_routes.py
asynchroza: Why are you dismissing unit tests especially in the context of an auth library? All of the cases you just explained can be validated with unit tests.

services/py-api/tests/integration_tests/test_jwt_utility.py
```diff
@@ -0,0 +1,82 @@
...
+ADDITIONAL_KEY_TEST_PAYLOAD = JwtUserData(
+    sub="alksdjflksd987fsidjf98sduf", is_admin=True, team_id="slkdjflkasjdflkjsdlkj", exp=sufficient_expiration_time, name="abcdefgji"  # type: ignore[typeddict-unknown-key]
+)
+
+EXPIRED_TEST_PAYLOAD = JwtUserData(
+    sub="alksdjflksd987fsidjf98sduf", is_admin=True, team_id="slkdjflkasjdflkjsdlkj", exp=instant_expiration_time
+)
+
+
+@patch.dict("os.environ", {"SECRET_KEY": "abcdefghijklmnopqrst"})
```
iobreshkov-qb, line 28: Is this patched correctly and why do we need to patch it? Also, as this is repeated maybe extract it as a variable.

Isak-Bego: This type of patching works because if you see the test passes when you try encoding/decoding with different keys. I do this so that we dont have to include the SECRET_KEY as a secret in github workflows so that the tests pass within the github actions virtual machines where are repository is mounted and tests are run

services/py-api/src/server/routes/utility_routes.py
Isak-Bego: Fixed in 2cf34e41919919fce21b30e78d7ee0dfaac03ca9

services/py-api/tests/integration_tests/test_jwt_utility.py
asynchroza: Correct

services/web/src/App.tsx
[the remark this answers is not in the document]
Isak-Bego: Fixed in f20e37676788f608c0c4dc31dfea05cd99f0deab

services/web/src/App.tsx
```diff
@@ -22,10 +22,6 @@ const RADIO_OPTIONS = [
     { label: 'Nothing', value: 'none' },
 ];
 
-import { startPerformanceTools } from 'reactscape-testing-tools';
-
-startPerformanceTools();
-
 function App() {
```
IvanObreshkov, line 25: No more performance upgrades :((((((((((((((((

services/py-api/src/utils.py
```diff
@@ -43,3 +48,42 @@ async def stop(self) -> None:
...
+            # this will help us see if all the fields that we have defined in the schema are present in the decoded
+            # jwt token
+            for key in schema.__annotations__.keys():
+                if key not in decoded_token:
+                    return Err("The decoded token does not correspond with the provided schema.")
+
+            # We check both sides since we dont know which of the dictionaries is bigger than the other
+            for key in decoded_token:
+                if key not in schema.__annotations__.keys():
+                    return Err("The decoded token does not correspond with the provided schema.")
```
IvanObreshkov, line 76: are we keeping the str errors, or should we return custom ones, as @asynchroza mentioned?

Isak-Bego: I just talked to @asynchroza and he thinks that it is better to keep them like this .

IvanObreshkov (approved):

PR #785, opened 2024-11-16
Installed jest and created a sample test #779

### LINKED ISSUE #779: Setup Jest Library
## Brief description:
Jest is a library for writing tests. Currenlty we don't have it setup. And you will have to do that.
## How to achieve it:
Go through Jest documentation and see how to set it up: [Jest Docs](https://jestjs.io/)

Pastedko:
Setup jest library and created a sample test in `/jest-test`

resolves #779 

## How to verify:

- `npm install`
- `npm test`


services/web/src/jest-test/Jest.test.tsx
```diff
@@ -0,0 +1,25 @@
+const renderer = require('react-test-renderer');
```
Aleks1908, line 1: Please rework the test to use @testing-library/react. The current test uses react-test-renderer and manually triggers callbacks, which can be simplified with @testing-library/react.

services/web/src/jest-test/Jest.tsx
```diff
@@ -0,0 +1,28 @@
+import React, { useState } from 'react';
```
Aleks1908, line 1: We are leaving this as of now so that the others have some template when they start writing their tests. But it should be removed later down the line

services/web/src/jest-test/__snapshots__/Jest.test.tsx.snap
```diff
@@ -0,0 +1,34 @@
+// Jest Snapshot v1, https://goo.gl/fbAQLP
```
Aleks1908, line 1: Same for this

Aleks1908 (changes requested):
You have to resolve the conflicts

services/web/src/jest-test/Jest.test.tsx
Pastedko: fixed in 961ed7d27f063f29b78bec23e10ef3c792dd7a15

Pastedko:
Also configured it to work with husky

Aleks1908 (approved):
LGTM

PR #790, opened 2024-11-17
#778 Implement radio button component

### LINKED ISSUE #778: [Internal lib] Research and implement a reusable radio button component
## Brief description:
We are going to be making an internal library for form elements. Research how you can do that based on the documentation of [shadcn](https://ui.shadcn.com/docs/components/radio-group) and [React-hook-form](https://react-hook-form.com/). The general idea is for these components to be fully reusable and to accept props for everything. So that we can not only use them in the HackAUBG page but in the admin panel in the future. You would also have to write tests with all the edge cases to verify that the radio buttons are working correctly. If you have any questions don't hesitate to ask @Aleks1908

MagdalenaZh:
Implemented a reusable radio button component using shadcn and made it compatible with React Hook Form.

resolves #778


services/web/package.json
```diff
@@ -12,6 +12,10 @@
   "dependencies": {
     "@hookform/resolvers": "^3.9.1",
     "@radix-ui/react-accordion": "^1.2.0",
+    "@radix-ui/react-label": "^2.1.0",
```
Aleks1908, line 15: These correspond to the components we have downloaded from shadcn. But we are not using all of them. Investigate which ones are not used and follow shadcn documentation on how to remove them.

services/web/src/App.tsx
```diff
@@ -1,42 +1,50 @@
-import { useState } from 'react';
-import reactLogo from './assets/react.svg';
-import viteLogo from '/vite.svg';
-import './App.css';
-import { QueryClient, QueryClientProvider } from '@tanstack/react-query';
-import { Accordion, AccordionContent, AccordionItem, AccordionTrigger } from '@radix-ui/react-accordion';
+'use client';
```
Aleks1908, line 1: This is not a `Next` project. And `use client` should not be here.

services/web/src/App.tsx
```diff
@@ -1,42 +1,50 @@
...
 
-import MockDataComponent from './website/mockComponent';
```
Aleks1908, line 8: Recover the old App.tsx and add the input. I had put a sample request component previously which will be useful later for you to see how it works.

services/web/src/index.css
```diff
@@ -72,7 +72,7 @@ button:focus-visible {
 }
```
Aleks1908, line 72: I assume these are linter changes. Investigate if everything is ok after your changes.

services/web/src/internal_library/radioComponent.tsx
```diff
@@ -0,0 +1,54 @@
+import { Control, FieldValues, Path } from 'react-hook-form';
```
Aleks1908, line 1: I wont go over the styling and props right now because I know its WIP

Aleks1908 (changes requested):

services/web/src/internal_library/radioComponent.tsx
```diff
@@ -0,0 +1,54 @@
...
+import { FormControl, FormField, FormItem, FormLabel, FormMessage } from '@/components/ui/form';
+import { RadioGroup, RadioGroupItem } from '@/components/ui/radio-group';
+
+interface RadioButtonProps<T extends FieldValues> {
```
Aleks1908, line 5: Make this a type instead of interface

services/web/src/App.tsx
```diff
@@ -1,16 +1,40 @@
...
+import reactLogo from './assets/react.svg';
+import viteLogo from '/vite.svg';
+import { useState } from 'react';
+
+const FormSchema = z.object({
+    notificationPreference: z.string().min(1, { message: 'GRESHKA' }),
+});
 
-function App() {
+export function App() {
```
Aleks1908, line 19: You should not be exporting App

services/web/src/App.tsx
```diff
@@ -1,16 +1,40 @@
...
+        defaultValues: {
+            notificationPreference: '',
+        },
+    });
+
+    const onSubmit = (data: z.infer<typeof FormSchema>) => {
+        console.log(data);
+    };
+
+    const radioOptions = [
```
Aleks1908, line 31: 1. Move `radioOptions` constant outside of the component, because this constant is not used in any hooks. We do this because React will recreate it on each rerender. 

2. For constants that are defined outside of the component we are going to stick to using: SCREAMING_SNAKE_CASE

Aleks1908 (changes requested):

services/web/src/internal_library/radioComponent.tsx
```diff
@@ -0,0 +1,58 @@
+import { Control, FieldValues, Path } from 'react-hook-form';
```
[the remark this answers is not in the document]
asynchroza: Ah damn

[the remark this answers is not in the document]
Aleks1908: Yup, I agree, but we will find a better name for the lib first cuz `internal-library` is too boring, thats why i didnt leave comment

services/web/src/internal_library/radioComponent.tsx
```diff
@@ -0,0 +1,58 @@
...
+import { FormControl, FormField, FormItem, FormLabel, FormMessage } from '@/components/ui/form';
+import { RadioGroup, RadioGroupItem } from '@/components/ui/radio-group';
+import { cn } from '@/lib/utils';
+
+type RadioButtonProps<T extends FieldValues> = {
+    control: Control<T>;
+    name: Path<T>;
+    options: { label: string; value: string }[];
```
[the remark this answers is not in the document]
asynchroza: You don't mean infer. You mean use.

```jsx
options: RadioButtonOptions
```

Inferring in JS is the following:

```jsx
const variable = "something"; // now if you hover over variable, language server tells you it's a string
```

and in typescript -> https://www.typescriptlang.org/docs/handbook/2/conditional-types.html#inferring-within-conditional-types

services/web/src/internal_library/radioComponent.tsx
```diff
@@ -0,0 +1,58 @@
...
+                        >
+                            {options.map((option) => (
+                                <FormItem
+                                    key={option.value}
+                                    className={cn('flex items-center space-x-3 space-y-0', formItemClassName)}
+                                >
+                                    <FormControl>
+                                        <RadioGroupItem value={option.value} className={cn('', radioGroupClassName)} />
+                                    </FormControl>
+                                    <FormLabel className={cn('', labelClassName)}>{option.label}</FormLabel>{' '}
```
asynchroza, line 48: You don't need `cn` here. Just pass labelClassName.

services/web/src/internal_library/radioComponent.tsx
```diff
@@ -0,0 +1,58 @@
...
+                            value={field.value}
+                            className={cn('', radioGroupClassName)}
+                        >
+                            {options.map((option) => (
+                                <FormItem
+                                    key={option.value}
+                                    className={cn('flex items-center space-x-3 space-y-0', formItemClassName)}
+                                >
+                                    <FormControl>
+                                        <RadioGroupItem value={option.value} className={cn('', radioGroupClassName)} />
```
asynchroza, line 46: You don't need cn here. Just pass radioGroupClassName.

services/web/src/App.tsx
asynchroza: It appears this is mostly for testing so that's a nit

asynchroza: But it's good if you adhere to what Aleks is talking about here. That's industry standard.

asynchroza: > We do this because React will recreate it on each rerender.

Essentially, React components are simple functions. If React rerenders the component, it means that the code inside the component is run once again. 

That's why if you have a state (useState) and a console.log inside a component, every time the state changes -> a rerender would be triggered and the log will be displayed in the console.

You can prevent the log from getting printed by putting inside a useEffect with an empty dependency array. But if the code is not wrapped within anything React related, it will always be rerun.

This means that any definition will be essentially reinitialized.

const something = {}; // something will be created every time the component rerenders

Why is this bad? JavaScript is a garbage collected language. Garbage collection is a really expensive process. Every time you create a new object and then stop using it, the garbage collector should mark the object as unused and collect it. Why keep creating new objects when you can just bring the object outside of the component if you know it's value is going to be constant anyways?

I hope this made sense. Feel free to ask any questions.

Aleks1908: Thank you for the clarifications

asynchroza:
Really good job on this one! Thanks to you both @Aleks1908 @MagdalenaZh

services/web/src/internal_library/RadioComponent/RadioComponent.tsx
```diff
@@ -0,0 +1,67 @@
...
+            control={control}
+            name={name}
+            render={({ field, fieldState: { error } }) => (
+                <FormItem className={cn('space-y-3', groupClassName)}>
+                    {groupLabel && <FormLabel>{groupLabel}</FormLabel>}
+                    <FormControl>
+                        <RadioGroup
+                            onValueChange={field.onChange}
+                            value={field.value}
+                            className={cn('', radioGroupClassName)}
```
Aleks1908, line 43: fix this

Aleks1908 (approved):
LGTM

PR #791, opened 2024-11-18
Remove iceberg class impl

IvanObreshkov:
@Isak-Bego 
It's generally a bad idea to test private methods not only because they are private and because you should test them via the public ones, but also because you are becoming too tightly coupled with a certain class's implementation. Private methods are by design supposed to hide what the end user of the API shouldn't be touching. This is also why interfaces are a thing - we only use the interface's methods, and any implementation which has private methods is effectively hidden from us. This way, if someone changes a class's inner workings, you can rely on the fact that you won't have to change anything about the way you call that class. If your test depends on a private method, you will need to change your test every time that this private method changes. What if it gets completely removed? 😃 

What other tests also use this method? You will have to refactor them as well. You get the same issue which Liskov substitution was designed to solve. 

That's why I told you to rewrite the tests. When this PR is merged, people are going to have a ton of conflicts, particularly @KDimitrov333, so help them with resolving.


services/py-api/src/database/repository/teams_repository.py
```diff
@@ -17,7 +17,7 @@
 
 class TeamsRepository(CRUDRepository):
     MAX_NUMBER_OF_TEAM_MEMBERS: Final[int] = 6
-    MAX_NUMBER_OF_TEAMS_IN_HACKATHON: Final[int] = 12
+    MAX_NUMBER_OF_VERIFIED_TEAMS_IN_HACKATHON: Final[int] = 12
```
Isak-Bego, line 20: Nice addition!

services/py-api/src/server/routes/participants_routes.py
```diff
@@ -30,16 +31,22 @@ def _tx_manager(db_manager: DB_MANAGER) -> TransactionManager:
...
 
-def _service(
+def _h_service(
     p_repo: ParticipantsRepository = Depends(_p_repo),
     t_repo: TeamsRepository = Depends(_t_repo),
     tx_manager: TransactionManager = Depends(_tx_manager),
-) -> ParticipantService:
-    return ParticipantService(p_repo, t_repo, tx_manager)
+) -> HackathonService:
+    return HackathonService(p_repo, t_repo, tx_manager)
```
Isak-Bego, line 39: Question: Does this make sure that we only have an instance of each Class running? What is the purpose?

services/py-api/src/server/routes/participants_routes.py
```diff
@@ -30,16 +31,22 @@ def _tx_manager(db_manager: DB_MANAGER) -> TransactionManager:
...
-def _handler(service: ParticipantService = Depends(_service)) -> ParticipantHandlers:
-    return ParticipantHandlers(service)
+def _p_reg_service(
+    h_service: HackathonService = Depends(_h_service),
+) -> ParticipantRegistrationService:
+    return ParticipantRegistrationService(h_service)
+
+
+def _handler(p_service: ParticipantRegistrationService = Depends(_p_reg_service)) -> ParticipantHandlers:
+    return ParticipantHandlers(p_service)
```
Isak-Bego, line 49: Nice dependency injection pattern between the 3 layers!

services/py-api/src/service/hackathon_service.py
```diff
@@ -25,10 +26,13 @@ def __init__(
...
 
-    async def _create_participant_and_team_in_transaction(
+    async def _create_participant_and_team_in_transaction_callback(
         self, input_data: ParticipantRequestBody, session: Optional[AsyncIOMotorClientSession] = None
     ) -> Result[Tuple[Participant, Team], DuplicateEmailError | DuplicateTeamNameError | Exception]:
-
+        """
+        This method is intended to be passed as the `callback` argument to the `TransactionManager.with_transaction(...)`
+        function.
+        """
```
Isak-Bego, line 35: Nice clarification comment!

services/py-api/tests/unit_tests/conftest.py
```diff
@@ -54,6 +56,9 @@ def close_side_effect() -> Optional[Err[str]]:
 
 @pytest.fixture
 def participant_repo_mock() -> Mock:
+    """This is a mock obj of ParticipantsRepository. To change the return values of its methods use:
+    `participant_repo_mock.method_name.return_value=some_return_value`"""
+
```
Isak-Bego, line 61: Nice clarification! Question for you: Although we are assigning these methods Mock objects, do these object contain the implementation of the original methods, or you can only bind them with return values?

services/py-api/tests/unit_tests/conftest.py
```diff
@@ -62,11 +67,16 @@ def participant_repo_mock() -> Mock:
     participant_repo.create = AsyncMock()
     participant_repo.delete = AsyncMock()
 
-    return participant_repo
+    yield participant_repo
+
+    participant_repo.reset_mock()
```
Isak-Bego, line 72: Question for the way I understand it: So basically here were reset the mock so that we have a clear mock object from one test to the other right?

IvanObreshkov: Yea it's a good practice, I believe I have put a comment somewhere up the code.

services/py-api/tests/unit_tests/test_handlers/test_participant_handlers.py
```diff
@@ -11,28 +11,22 @@
...
 @pytest.fixture
-def participant_service_mock() -> Mock:
-    return Mock(ParticipantService, register_admin_participant=AsyncMock())
-
-
-@pytest.fixture
-def participant_handlers(participant_service_mock: Mock) -> ParticipantHandlers:
-    return ParticipantHandlers(participant_service_mock)
+def participant_handlers(participant_registration_service_mock: Mock) -> ParticipantHandlers:
+    return ParticipantHandlers(participant_registration_service_mock)
```
Isak-Bego, line 18: Add a comment for where we get this argument because it is not imported. How does conftest.py work throughout the whole testing directory. Maybe add some more docs about how fixtures work

services/py-api/tests/unit_tests/conftest.py
```diff
@@ -13,6 +13,8 @@
 from src.database.repository.teams_repository import TeamsRepository
 from src.database.transaction_manager import TransactionManager
 from src.server.schemas.request_schemas.schemas import ParticipantRequestBody
+from src.service.hackathon_service import HackathonService
+from src.service.participants_registration_service import ParticipantRegistrationService
```
Isak-Bego, line 19: Somewhere here we can do a very brief explaination on how MagicMock(), AsyncMock() and just Mock() work, with links to the docs

Isak-Bego: We can also say a word or 2 about conftest.py and its role in pytest

services/py-api/tests/unit_tests/conftest.py
```diff
@@ -62,11 +67,16 @@ def participant_repo_mock() -> Mock:
...
+    yield participant_repo
+
+    participant_repo.reset_mock()
 
 
 @pytest.fixture
 def team_repo_mock() -> Mock:
+    """This is a mock obj of TeamsRepository. To change the return values of its methods use:
+    `team_repo_mock.method_name.return_value=some_return_value`"""
+
```
Isak-Bego, line 79: I absolutely love these comments! Good job mate!

services/py-api/tests/unit_tests/test_service_layer/test_participant_registration_service.py
```diff
@@ -0,0 +1,172 @@
...
+    hackathon_service_mock.create_participant_and_team_in_transaction.return_value = Err(
+        DuplicateTeamNameError(mock_input_data.team_name)
+    )
+
+    # Call the function under test
+    result = await p_reg_service.register_admin_participant(mock_input_data)
+
+    # Check that the result is an `Err` with `DuplicateTeamNameError`
+    assert isinstance(result, Err)
+    assert isinstance(result.err_value, DuplicateTeamNameError)
```
Isak-Bego, line 73: The comments are so on point. and I love the strategy that we are following for testing, now it actually makes sense. 🚀

Isak-Bego (changes requested):
Just some suggestions for adding more docs here and there. Everything else is very nicely done 🚀

services/py-api/src/server/routes/participants_routes.py
Isak-Bego: Yeah its fine! Thank you!

services/py-api/tests/unit_tests/conftest.py
IvanObreshkov: Mock objects are mock objects, meaning fake. The idea is that you don't want to call the original method. So, no, they do not contain the implementation of the original methods. they are just fake properties/methods called during the execution of the test. And you can make them return whatever you want or add side effects, for example, throwing exceptions.

services/py-api/tests/unit_tests/test_handlers/test_participant_handlers.py
IvanObreshkov: https://github.com/AUBGTheHUB/monolith/blob/main/services/py-api/tests/unit_tests/README.md

services/py-api/tests/unit_tests/conftest.py
IvanObreshkov: https://github.com/AUBGTheHUB/monolith/blob/main/services/py-api/tests/unit_tests/README.md

services/py-api/tests/unit_tests/conftest.py
IvanObreshkov: Fixed in 856b2d8a134a84cab904ff83ee54d0dea708d92f 
I removed it as it was redundant due to fixtures having `scope=function` by default. https://docs.pytest.org/en/stable/how-to/fixtures.html#fixture-scopes

services/py-api/tests/unit_tests/test_handlers/test_participant_handlers.py
IvanObreshkov: I updated the README in 2973c46fb277bc549a219ceb1d03bcceddbbb2f1

services/py-api/tests/unit_tests/conftest.py
IvanObreshkov: Updated the README in 2973c46fb277bc549a219ceb1d03bcceddbbb2f1

PR #792, opened 2024-11-18
Add retry to single db operations

IvanObreshkov:
In the future, we may need to use single db operations. As Mongo Atlas is a managed db and is deployed in AWS Frankfurt (eu-central-1), it is also clustered. Having a retry mechanism when working with remote systems is a good practice to increase reliability.

### DIFF
--- a/services/py-api/src/database/db_manager.py
+++ b/services/py-api/src/database/db_manager.py
@@ -5,7 +5,7 @@ from typing import Annotated, Callable, Awaitable, Any
 
 from fastapi import Depends
 from motor.motor_asyncio import AsyncIOMotorClient, AsyncIOMotorCollection
-from pymongo import MongoClient
+from pymongo import MongoClient, helpers_shared
 from pymongo.errors import ConnectionFailure, OperationFailure, ConfigurationError, PyMongoError
 from result import Err
 from structlog.stdlib import get_logger
@@ -68,16 +68,27 @@ class DatabaseManager(metaclass=SingletonMeta):
         return self._client[self._DB_NAME][collection_name]
 
     @staticmethod
-    async def retry_db_operation[T](db_operation: Callable[..., Awaitable[T]], *args: Any, **kwargs: Any) -> T:
+    async def retry_db_operation[
+        T
+    ](db_operation: Callable[..., Awaitable[T]], is_read_operation: bool, *args: Any, **kwargs: Any) -> T:
         """
-        If there was a TransientTransactionError during execution, retries the db operation with exponential backoff up
-        to max_retries times. Having an exponential backoff increases our chances of success if there was a temporary
-        network issue.
+        By default, MongoDB retries read and write operations, eligible for retry, only once. To increase the
+        reliability of our system we retry these operations additional two times. With this method a read or write
+        operation, eligible for retry, is retried in total of 3 times.
+
+        To learn more:
+        https://www.mongodb.com/docs/manual/core/retryable-writes/
+
+        https://github.com/mongodb/specifications/blob/master/source/retryable-writes/retryable-writes.md
+
+        https://www.mongodb.com/docs/manual/core/retryable-reads/
+
+        https://github.com/mongodb/specifications/blob/master/source/retryable-reads/retryable-reads.md
 
         Raises:
             PyMongoError
         """
-        max_retries = 3
+        max_retries = 2
         delay = 0.2  # initial delay in seconds (200 milliseconds)
 
         # range is exclusive that's why we do max_retries + 1
@@ -85,13 +96,28 @@ class DatabaseManager(metaclass=SingletonMeta):
             try:
                 return await db_operation(*args, **kwargs)
             except PyMongoError as exc:
-                if exc.has_error_label("TransientTransactionError"):
-                    LOG.debug("Retrying database operation. Retry {}".format(retry))
+                # Retryable write error
+                # https://github.com/mongodb/specifications/blob/master/source/retryable-writes/retryable-writes.md#retryablewriteerror-labels
+                if exc.has_error_label("RetryableWriteError") and is_read_operation is False:
+                    LOG.debug("Retrying database write operation. Retry {}".format(retry))
                     await sleep(delay)
                     delay *= 2  # exponential backoff
                     continue
 
-                # If the exception it's a non-retryable error or a transactional error which didn't manage to propagate
+                # Retryable read error
+                # https://github.com/mongodb/mongo-python-driver/blob/2d21035396f63437176965cc8f157505189f2f08/pymongo/asynchronous/mongo_client.py#L2590-L2603
+                if isinstance(exc, (ConnectionFailure, OperationFailure)) and is_read_operation:
+                    # ConnectionFailures do not supply a code property that's why None is passed as a third argument
+                    # If the read error is not retryable re-raise it
+                    if getattr(exc, "code", None) not in helpers_shared._RETRYABLE_ERROR_CODES:
+                        raise exc
+
+                    LOG.debug("Retrying database read operation. Retry {}".format(retry))
+                    await sleep(delay)
+                    delay *= 2  # exponential backoff
+                    continue
+
+                # If the exception it's a non-retryable error or other error which didn't manage to propagate
                 # re-raise it in order to be caught on an upper level
                 raise exc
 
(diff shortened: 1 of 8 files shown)


Isak-Bego:
Can't we use the transaction logic and just create another callback that only adds a participant. I think that would be much easier

Isak-Bego (commented):
Can't we use the transaction logic and just create another callback that only adds a participant. I think that would be much easier

IvanObreshkov:
> Can't we use the transaction logic and just create another callback that only adds a participant. I think that would be much easier

Can you expand on this? I didn't want to use the `_retry_tx` method as it is a provate method. About the `with_transaction` as @asynchroza has mentioned before and also according to [this stack overflow thread](https://stackoverflow.com/questions/1171749/what-does-a-transaction-around-a-single-statement-do) having a transaction over a single db operation is useless. Firstly, single db operations are atomic, meaning they either succeed or fail. And secondly transactions create overhead which we would like to avoid when not needed.

If you got misled by the log message, sorry I will fix it.

services/py-api/src/database/db_manager.py
```diff
@@ -66,6 +67,35 @@ def get_collection(self, collection_name: str) -> AsyncIOMotorCollection:
...
+        """
+        If there was a TransientTransactionError during execution, retries the db operation with exponential backoff up
+        to max_retries times. Having an exponential backoff increases our chances of success if there was a temporary
+        network issue.
+
+        Raises:
+            PyMongoError
+        """
+        max_retries = 3
+        delay = 1  # initial delay in seconds
```
asynchroza, line 81: Isn't this too big of a delay? Especially, if you're doing an exponential backoff this seems to be a pretty high value.

services/py-api/src/database/db_manager.py
```diff
@@ -66,6 +67,35 @@ def get_collection(self, collection_name: str) -> AsyncIOMotorCollection:
...
+        # range is exclusive that's why we do max_retries + 1
+        for retry in range(1, max_retries + 1):
+            try:
+                return await db_operation(*args, **kwargs)
+            except PyMongoError as exc:
+                if exc.has_error_label("TransientTransactionError"):
+                    LOG.debug("Retrying transaction retry {}".format(retry))
+                    await sleep(delay)
+                    delay *= 2  # exponential backoff
+                    continue
```
asynchroza, line 92: I think it would be better if you wrap the exception raise in an else block and not use `continue`.

asynchroza: Oh, wait. I'm seeing where my logic fails.

services/py-api/src/database/db_manager.py
```diff
@@ -66,6 +67,35 @@ def get_collection(self, collection_name: str) -> AsyncIOMotorCollection:
...
+
+                # If the exception it's a non-retryable error re-raise it in order to be caught on an upper level
```
asynchroza, line 94: Update the comment to include that it's either a non-retryable error or a transactional error which didn't manage to propagate

services/py-api/src/database/db_manager.py
```diff
@@ -66,6 +67,35 @@ def get_collection(self, collection_name: str) -> AsyncIOMotorCollection:
...
+                raise exc
+
+        raise PyMongoError("Database operation failed after maximum retries")
```
asynchroza, line 97: This error can be misleading. If it's a non-retryable error, the operation failed without retrying.

services/py-api/src/database/repository/participants_repository.py
```diff
@@ -40,20 +41,34 @@ async def create(
         session: Optional[AsyncIOMotorClientSession] = None,
         **kwargs: Dict[str, Any]
     ) -> Result[Participant, DuplicateEmailError | Exception]:
+
+        participant = Participant(
+            name=input_data.name,
+            email=input_data.email,
+            is_admin=input_data.is_admin,
+            # If the team_id is passed as a kwarg the participant will be inserted in the given team
```
asynchroza, line 49: This comment should be added as a docstring. This is really important.

services/py-api/src/database/repository/teams_repository.py
```diff
@@ -29,10 +30,21 @@ async def create(
         **kwargs: Dict[str, Any]
     ) -> Result[Team, DuplicateTeamNameError | Exception]:
 
+        team = Team(name=input_data.team_name)
+
+        async def db_operation() -> None:
+            await self._collection.insert_one(document=team.dump_as_mongo_db_document(), session=session)
```
asynchroza, line 36: Wouldn't it have made sense to actually create to have an abstraction on top of collection which accepts the actual pydantic object (or whatever you're using).

iobreshkov-qb: We are using custom data classes due to serialization problems with the ObejctID. You can check them [here](https://github.com/AUBGTheHUB/monolith/tree/main/services/py-api/src/database/model). TBH, I never thought about such abstraction, I thought the repository layer was good enough, and creating the types inline was just fine. We could think about your idea, but it will require a lot of changes.

asynchroza: I'm too lazy to explain why this is so important but the TL;DR - Better DevEx, less prone to errors because dump happens in the base concrete class of `MongoCollection` which is extended by all of the collections -> you get to only have to deal with your custom model objects and not manually dump them. Do not consider this a mandatory change by any means. I'm just pointing it out.

services/py-api/src/database/repository/participants_repository.py
IvanObreshkov: Fixed in c3390a1c03923cd8ec9f09bcc4f47a9a0ac2942d

services/py-api/src/database/db_manager.py
IvanObreshkov: fixed in 43821673891e234b5d62551c4ee093940a082253 and c6a67ae107d5b1a2fc4edfcb3725db905d2acdce

services/py-api/src/database/db_manager.py
IvanObreshkov: If the exception is a non-retryable error it is raised directly and we don't even get to `raise PyMongoError("Database operation failed after maximum retries")`. If any other error different from `PyMongoError` is raised, we again don't hit this line, and the error is caught by an upper-level try-catch block

services/py-api/src/database/db_manager.py
```diff
@@ -66,6 +67,35 @@ def get_collection(self, collection_name: str) -> AsyncIOMotorCollection:
         # https://pymongo.readthedocs.io/en/stable/tutorial.html#getting-a-collection
         return self._client[self._DB_NAME][collection_name]
 
+    @staticmethod
+    async def retry_db_operation[T](db_operation: Callable[..., Awaitable[T]], *args: Any, **kwargs: Any) -> T:
```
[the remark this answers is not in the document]
IvanObreshkov: I am open to suggestions :D

services/py-api/src/database/db_manager.py
IvanObreshkov: Fixed in 062077d0b7deca169495ff24160e440924375f68. Keep in mind that the logic in `db_manager.py` might change as I am currently working on it, and the requsted change is made in the `transaction_manager.py`

services/py-api/src/database/repository/participants_repository.py
IvanObreshkov: It is added to the base "create" method docs:
image

services/py-api/src/database/repository/participants_repository.py
```diff
@@ -38,15 +39,28 @@ async def delete(
...
+
+        try:
+            LOG.debug("Deleting participant...", participant_obj_id=obj_id)
+
+            # The `TransactionManager.with_transaction` method provides the session and has a built-in retry
+            # mechanism
+            if session:
+                result = await db_operation()
+            else:
+                result = await self._db_manager.retry_db_operation(db_operation, is_read_operation=False)
```
Isak-Bego, line 64: This approach requires us to remember to always write the retry mechanism in the end and define the operations inside a db_operation(), which could not be the best DEV experience we want to offer and could be error prone. Did you consider writing a decorator for this approach? It seems like it is a perfect candidate to be a decorator, because it includes a nested function that is always used by the retry mechanism. If it is possible, I think this would be a better approach.

IvanObreshkov:
Not important as of now, Mongo has a one time retry which we trust.

PR #795, opened 2024-11-20
Do some cleanup

IvanObreshkov:
PLS test if the gum tool works and the setup in general


Makefile
```diff
@@ -1,60 +1,24 @@
...
-.PHONY: install-env
-install-env:
-	cp -n .env.sample .env; \
-	ln -sf ${PWD}/.env ${PWD}/services/web/.env.development; \
-	ln -sf ${PWD}/.env ${PWD}/services/api/.env; \
-	ln -sf ${PWD}/.env ${PWD}/services/url_shortener/.env; \
-	ln -sf ${PWD}/.env ${PWD}/services/py-api/.env; \
-	ln -sf ${PWD}/.env ${PWD}/services/questionnaire;
-
-.PHONY: run-py-api
```
Isak-Bego, line 41: Since we are not using this anymore, we should remove it from the postCreateCommand.sh in the .devcontainer folder

Makefile
```diff
@@ -1,60 +1,24 @@
...
-.PHONY: install-signed-certs
-install-signed-certs:
-	cp data/certs/local.crt data/certs/devenv.crt
-	cp data/certs/local.key data/certs/devenv.key
-
-.PHONY: run-rust-api
 run-rust-api:
 	cd services/url_shortener && make watch
 
-.PHONY: run-svelte-quest
```
Isak-Bego, line 54: Are we sure we want to get rid of the PHONY? 
I think this https://stackoverflow.com/questions/2145590/what-is-the-purpose-of-phony-in-a-makefile makes a good point of why we want to keep it.

Isak-Bego:
> PLS test if the gum tool works and the setup in general

image
Gives this error so I guess lets just remove it like @mbozhilov-qb suggests.

Makefile
IvanObreshkov: Do we have files named like the Make targets?

Isak-Bego: Just to be sure man. It doesn't hurt, since we are running commands with it it would be more correct to have the phonies. We don't know how the project is going to evolve in the future

Isak-Bego (approved):
LGTM

PR #817, opened 2024-12-11
#807 added admin participant verification

### LINKED ISSUE #807: feature: implement the verification of admin participant
## Brief description:
STORY: 
>"You just registered your team in the hackathon, and you receive an verification email from the Hub. Once you click in the verification link that is provided to you, you verify yourself and secure a spot in the hackathon"

We receive the JWT on our verification endpoint **/hackathon/participants/verify?jwt_token=sdnflksduf09s8d02ojfiw029**. We decode the token. If the decoding is successful (the token has not expired nor has it been altered by a third party) we examine the decoded token. This issue concerns only the tokens whose payloads have is_admin=True (since we are dealing with admin participants). After performing the necessary checks, you update the team and participant(admin) documents in the database by switching email_verified=True and is_verified=True respectively for the participant document and team document (you can check the database models to make sure). In this case these database operations should be performed in a transaction using the same logic as we use for the creation of the admin participant. 

## How to achieve it:
1. In the verification_handler.py decode the token and create a check for the is_admin field of the payload. If is_admin is true you continue with the rest of your logic in the controller by calling verify_admin_participant() method of the verification service class that will be created in #805. 
2. Follow this logic in the controller: 
image
Here is a link to the API design schema: https://excalidraw.com/#json=qJbg3VyVnKP_jOAlDq2Kf,BileoGllBKsbeQOf8KydlQ
3. Keep in mind that for now we are not doing the sending of emails since we don't have the logic ready.

Bubomira:
#807 Verify Admin Participant

##
1) added route for participant verification
2) added verification handler:
--checks if the jwt token is valid
--if it is it calls verify_admin_participant in participant_verification_service
3) verify_admin_participant in participant_verification_service
4) _verify_admin_participant_and_team_callback() and verify_admin_participant_and_team_in_transaction() in hackaton service
5) update methods for both the participants and teams repositories
6) unit tests for
--update methods in team and participants repository
--verify_admin_participant_and_team_in_transaction in hackathon_service
--verify_admin_participant in participant_verification_service
--verify_participant (admin case) in verification_handler
7) integration tests

resolves #807


services/py-api/src/service/verification_service.py
IvanObreshkov, line 1: Do we need a separate service just for two wrapper methods? I agree that you have done a "separation of concern" and are adhering to the "S" from SOLID, but in this case, it creates a lot of boilerplates. We need to create a separate mock for this service when we are writing tests, also you are adding one more dependency which you have to instantiate and take care of. 

The logic in `VerificationService` is not so sophisticated that it requires a separate service layer on its own, as here you are just calling methods from the `HackathonService`.

`HackathonService` in our codebase is the place where all common logic and operations regarding the hackathon registration process are encapsulated.  Yes, `HackathonService` could be argued that it breaks "S" from SOLID, but I don't see it like that. `HackathonService` acts only as a container, and the methods in it adhere to the Single Responsibility principle.

services/py-api/src/service/hackathon_service.py
```diff
@@ -65,6 +65,24 @@ async def create_random_participant(
         # As when first created, the random participant is not assigned to a team we return the team as None
         return Ok((result.ok_value, None))
 
+    async def verify_admin_participant_and_team_in_transaction(self, admin_id: str) -> Any:
```
IvanObreshkov, line 68: Please return Concrete types, not Any. Here, the type could be `Result[Tuple[Participant, Team], ParticipantNotFoundError | TeamNotFoundError | Exception` 

Also didn't you get an error from MyPy? It is set to [strict](https://mypy.readthedocs.io/en/stable/command_line.html#cmdoption-mypy-strict) in its [config file](https://github.com/AUBGTheHUB/monolith/blob/main/services/py-api/.pre-commit-config.yaml): 
image

strict enables all optional flags including [--no-strict-optional](https://mypy.readthedocs.io/en/stable/command_line.html#cmdoption-mypy-no-strict-optional)

services/py-api/src/service/hackathon_service.py
```diff
@@ -65,6 +65,24 @@ async def create_random_participant(
...
+            obj_id=admin_id, input_data={"email_verified": True}, session=session
+        )
+        if is_err(result_admin):
+            return result_admin
+
+        result_team = await self._team_repo.update(obj_id=admin_id, input_data={"is_verified": True}, session=session)
+        if is_err(result_team):
+            return result_admin
+
+        return Ok(value=True)
```
IvanObreshkov, line 84: Why are you returning boolean here? Maybe return the updated Team and Participant documents.

services/py-api/src/service/hackathon_service.py
```diff
@@ -101,3 +119,6 @@ async def check_capacity_register_random_participant_case(self) -> bool:
 
         # Check against the hackathon capacity
         return number_ant_teams <= self._team_repo.MAX_NUMBER_OF_VERIFIED_TEAMS_IN_HACKATHON
+
+    async def check_if_participant_exists_in_by_id(self, object_id: str) -> bool:
+        return await self._participant_repo.check_if_participant_exists_by_id(obj_id=object_id)
```
IvanObreshkov, line 124: Not sure we need this method, more info is in the comments below.

services/py-api/src/server/routes/verification_routes.py
```diff
@@ -0,0 +1,50 @@
...
+    h_service: HackathonService = Depends(_h_service),
+) -> VerificationService:
+    return VerificationService(h_service)
+
+
+def _handler(v_service: VerificationService = Depends(_v_service)) -> VerificationHandlers:
+    return VerificationHandlers(v_service)
+
+
+@verification_router.get("", status_code=200)
```
IvanObreshkov, line 46: Shouldn't this be a `PATCH` method? According to [RFC 7231](https://datatracker.ietf.org/doc/html/rfc7231#section-4.3.1) "GET"  is used to retrieve resources, and here we are Updating a resource.

According to the same memo [PUT](https://datatracker.ietf.org/doc/html/rfc7231#section-4.3.4) should be used to replace or create a resource if the resource does not exist. In other words an UPSERT operation. 

Here, we are not replacing a whole resource; we are changing some of its fields, or in other words, we are doing "Partial content updates" and [the PATCH METHOD](https://datatracker.ietf.org/doc/html/rfc5789#section-2) should be used as stated in the section for "PUT".

services/py-api/src/database/repository/teams_repository.py
```diff
@@ -57,9 +58,15 @@ async def update(
...
-        raise NotImplementedError()
+        **kwargs: Dict[str, Any],
+    ) -> Result[Ok, Exception]:
+        try:
+            LOG.debug(f"Updating team with id {obj_id}")
+            await self._collection.find_one_and_update({"_id": ObjectId(obj_id)}, {"$set": input_data}, session=session)
+            return Ok(value=True)
+        except Exception as e:
+            LOG.error(f"Updating team with id {obj_id} failed due to err {e}")
+            return Err(e)
```
IvanObreshkov, line 69: Please check my comment in the `ParticipantRepository`

services/py-api/src/server/handlers/verification_handlers.py
```diff
@@ -0,0 +1,44 @@
...
+            response.status_code = status.HTTP_409_CONFLICT
+            return ErrResponse(error="Participant does not exist in database")
+
+        if jwt_payload.ok_value.get("is_admin"):
+            result = await self._verification_service.verify_admin_participant(participant_id)
+
+            if is_err(result):
+                if isinstance(result.err_value, HackathonCapacityExceededError):
+                    response.status_code = status.HTTP_409_CONFLICT
+                    return ErrResponse(error="Max hackathon capacity has been reached")
```
IvanObreshkov, line 36: New line pls. It improves readability.

services/py-api/src/database/repository/participants_repository.py
```diff
@@ -28,8 +29,14 @@ async def fetch_all(self) -> Result:
...
-    ) -> Result:
-        raise NotImplementedError()
+    ) -> Result[Ok, Exception]:
+        try:
+            LOG.debug(f"Updating participant with id {obj_id}")
+            await self._collection.find_one_and_update({"_id": ObjectId(obj_id)}, {"$set": input_data}, session=session)
+            return Ok(value=True)
+        except Exception as e:
+            LOG.error(f"Updating participant with id {obj_id} failed due to err {e}")
+            return Err(e)
```
IvanObreshkov, line 39: Please use [LOG.exception](https://www.structlog.org/en/stable/api.html#structlog.stdlib.BoundLogger.exception) so that we can preserve the stack trace in our log files. 

For more info: https://betterstack.com/community/guides/logging/structlog/#logging-exceptions-with-structlog
https://www.structlog.org/en/stable/api.html#structlog.processors.StackInfoRenderer

services/py-api/src/server/handlers/verification_handlers.py
```diff
@@ -0,0 +1,44 @@
...
+                response.status_code = 500
+                return ErrResponse(
+                    error="An unexpected error occurred during the verification of the admin participant"
+                )
+
+            return Response(content="Successfully verifyed admin participant")
+        #  Validation check for random participants
+        return ErrResponse(error="Not implemented yet")
```
IvanObreshkov, line 44: Ok for now.

services/py-api/src/server/handlers/verification_handlers.py
```diff
@@ -0,0 +1,44 @@
...
+            if is_err(result):
+                if isinstance(result.err_value, HackathonCapacityExceededError):
+                    response.status_code = status.HTTP_409_CONFLICT
+                    return ErrResponse(error="Max hackathon capacity has been reached")
+                response.status_code = 500
+                return ErrResponse(
+                    error="An unexpected error occurred during the verification of the admin participant"
+                )
+
+            return Response(content="Successfully verifyed admin participant")
```
IvanObreshkov, line 42: You have a typo; it should be "verified"

services/py-api/src/server/handlers/verification_handlers.py
```diff
@@ -0,0 +1,44 @@
...
+        if is_err(jwt_payload):
+            response.status_code = status.HTTP_401_UNAUTHORIZED
+            return ErrResponse(error=jwt_payload.err_value)
+
+        participant_id = jwt_payload.ok_value.get("sub")
+
+        participant_exists = await self._verification_service.verify_participant_exists(id=participant_id)
+        if not participant_exists:
+            response.status_code = status.HTTP_409_CONFLICT
+            return ErrResponse(error="Participant does not exist in database")
```
IvanObreshkov, line 28: A 409 status code for a non-existing resource? According to [MDN](https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/409) 409 indicates a conflict, meaning a duplicate resource. You should use the [404](https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/404) status code.

services/py-api/src/server/routes/verification_routes.py
```diff
@@ -0,0 +1,50 @@
...
+    t_repo: TeamsRepository = Depends(_t_repo),
+    tx_manager: TransactionManager = Depends(_tx_manager),
+) -> HackathonService:
+    return HackathonService(p_repo, t_repo, tx_manager)
+
+
+def _v_service(
+    h_service: HackathonService = Depends(_h_service),
+) -> VerificationService:
+    return VerificationService(h_service)
```
IvanObreshkov, line 39: FYI these will be moved to a shared location. Check up with @Isak-Bego on updates regarding his pr https://github.com/AUBGTheHUB/monolith/pull/813 as you will have merge conflicts

services/py-api/src/service/verification_service.py
```diff
@@ -0,0 +1,18 @@
...
+    def __init__(self, hackaton_service: HackathonService):
+        self._hackaton_service = hackaton_service
+
+    async def verify_participant_exists(self, id: str) -> bool:
+        return await self._hackaton_service.check_if_participant_exists_in_by_id(object_id=id)
+
+    async def verify_admin_participant(self, id: str) -> Any:
+        has_capacity = await self._hackaton_service.check_capacity_register_admin_participant_case()
+        if not has_capacity:
+            return Err(HackathonCapacityExceededError())
```
IvanObreshkov, line 17: No need for this check here as if the `update` methods in the repositories are updated to return `NotFound` Exceptions, `verify_admin_participant_and_team_callback` will terminate. In this way, you are doing one less query as it is taken care of by the database.

services/py-api/src/service/hackathon_service.py
```diff
@@ -65,6 +65,24 @@ async def create_random_participant(
...
         return Ok((result.ok_value, None))
 
+    async def verify_admin_participant_and_team_in_transaction(self, admin_id: str) -> Any:
+        return await self._tx_manager.with_transaction(self.verify_admin_participant_and_team_callback, admin_id)
+
+    async def verify_admin_participant_and_team_callback(
+        self, admin_id: str, session: Optional[AsyncIOMotorClientSession] = None
+    ) -> Any:
+        result_admin = await self._participant_repo.update(
+            obj_id=admin_id, input_data={"email_verified": True}, session=session
```
IvanObreshkov, line 75: image
Fix the type, please.

services/py-api/src/database/repository/participants_repository.py
```diff
@@ -22,15 +22,29 @@ def __init__(self, db_manager: DatabaseManager, collection_name: str) -> None:
...
-        raise NotImplementedError()
+    ) -> Result[Ok, Exception]:
+        try:
+            LOG.debug(f"Updating participant with id {obj_id}")
+            result = await self._collection.find_one_and_update(
+                {"_id": ObjectId(obj_id)}, {"$set": input_data}, session=session
+            )
+            if result:
+                LOG.debug(f"Successfully updated participant with id {obj_id}")
+                return Ok(result)
```
Isak-Bego, line 40: Return something like this: 
```Python
# The result is None when the participant with the specified ObjectId is not found
            if result:
                return Ok(Participant(id=obj_id, **result))
```

You should return a Participant Object inside of and Ok object. To find the object id and the rest of the input data rely on the `input_data` parameter

services/py-api/src/database/repository/participants_repository.py
```diff
@@ -22,15 +22,29 @@ def __init__(self, db_manager: DatabaseManager, collection_name: str) -> None:
...
+        return await self._collection.find_one({"_id": ObjectId(obj_id)})
 
     async def fetch_all(self) -> Result:
         raise NotImplementedError()
 
     async def update(
         self, obj_id: str, input_data: BaseModel, session: Optional[AsyncIOMotorClientSession] = None, **kwargs: Any
-    ) -> Result:
-        raise NotImplementedError()
+    ) -> Result[Ok, Exception]:
```
Isak-Bego, line 32: You have misunderstood the type def. 
Instead of what you have written it should have been: 
```Python
Result[Participant, ParticipantNotFoundError | Exception]
# The idea here is that in the first part you include the type of whatever you put
# inside Ok(...) and on the second part (after the comma) you include the types of whatever
# you put inside Err(...)
```

Isak-Bego: Please check the rest of the examples for clarification

services/py-api/src/database/repository/participants_repository.py
```diff
@@ -22,15 +22,29 @@ def __init__(self, db_manager: DatabaseManager, collection_name: str) -> None:
...
+            result = await self._collection.find_one_and_update(
+                {"_id": ObjectId(obj_id)}, {"$set": input_data}, session=session
+            )
+            if result:
+                LOG.debug(f"Successfully updated participant with id {obj_id}")
+                return Ok(result)
+            LOG.exception(f"No updated participants because participant with id {obj_id} was not found")
+            return Err(
+                ParticipantNotFoundError(f"No updated participants because participant with id {obj_id} was not found")
+            )
```
Isak-Bego, line 44: You don't have to put a message inside the ParticipantNotFoundError(). Simply leave it as: 
```Python
ParticipantNotFoundError()
```

services/py-api/src/database/repository/teams_repository.py
```diff
@@ -58,9 +58,21 @@ async def update(
...
+                {"_id": ObjectId(obj_id)}, {"$set": input_data}, session=session
+            )
+            if result:
+                LOG.debug(f"Successfully updated team with id {obj_id}")
+                return Ok(result)
+            LOG.exception(f"No updated teams because team with id {obj_id} was not found")
+            return Err(TeamNotFoundError(f"No updated teams because team with id {obj_id} was not found"))
+        except Exception as e:
+            LOG.exception(f"Updating team with id {obj_id} failed due to err {e}")
+            return Err(e)
```
Isak-Bego, line 75: The same comments that apply to the ParticipantsRepository apply here too. Instead you have objects like: Team(...) and TeamNotFoundError(...). Check the delete operation as an example.

services/py-api/src/server/routes/verification_routes.py
```diff
@@ -0,0 +1,43 @@
...
+def _tx_manager(db_manager: DB_MANAGER) -> TransactionManager:
+    return TransactionManager(db_manager)
+
+
+def _h_service(
+    p_repo: ParticipantsRepository = Depends(_p_repo),
+    t_repo: TeamsRepository = Depends(_t_repo),
+    tx_manager: TransactionManager = Depends(_tx_manager),
+) -> HackathonService:
+    return HackathonService(p_repo, t_repo, tx_manager)
```
Isak-Bego, line 32: All of these have moved to routes/dependency_factory.py. So you don't need to redefine them, just import them to this file

services/py-api/src/server/handlers/verification_handlers.py
```diff
@@ -0,0 +1,56 @@
...
+from starlette import status
+
+from src.utils import JwtUtility
+from src.server.schemas.response_schemas.schemas import ErrResponse
+from src.server.schemas.jwt_schemas.jwt_user_data_schema import JwtUserData
+
+
+class VerificationHandlers:
+    def __init__(self, hackaton_service: HackathonService) -> None:
+        self._hackaton_service = hackaton_service
```
Isak-Bego, line 14: Typo: Should be `hackathon` instead of hackaton

services/py-api/src/service/hackathon_service.py
```diff
@@ -107,6 +129,10 @@ async def check_capacity_register_random_participant_case(self) -> bool:
         # Check against the hackathon capacity
         return number_ant_teams <= self._team_repo.MAX_NUMBER_OF_VERIFIED_TEAMS_IN_HACKATHON
 
+    async def check_if_participant_exists_in_by_id(self, object_id: str) -> bool:
+        result = await self._participant_repo.fetch_by_id(obj_id=object_id)
+        return result is not None
+
```
Isak-Bego, line 135: You don't need this. Instead you can use something like: 

```Python
# Check if team still exists - Returns an error when it doesn't
        result = await self._team_repo.fetch_by_team_name(input_data.team_name)  # type: ignore
        if is_err(result):
            return result
```
which is inside of

```Python
    async def create_invite_link_participant(
        self, input_data: ParticipantRequestBody, decoded_jwt_token: JwtUserData
    ) -> Result[
        Tuple[Participant, None], DuplicateEmailError | TeamCapacityExceededError | TeamNotFoundError | Exception
    ]:

        # Check if team still exists - Returns an error when it doesn't
        result = await self._team_repo.fetch_by_team_name(input_data.team_name)  # type: ignore
        if is_err(result):
            return result

        # Check Team Capacity
        has_capacity = await self.check_team_capacity(decoded_jwt_token["team_id"])
        if not has_capacity:
            return Err(TeamCapacityExceededError())

        participant_result = await self._participant_repo.create(
            input_data, team_id=SerializableObjectId(decoded_jwt_token["team_id"]), email_verified=True
        )

        if is_err(participant_result):
            return participant_result

        # Return the new participant
        return Ok((participant_result.ok_value, None))

```

You can use the repository method directly as a check. You don't need a separate method in the Hackathon Service class.

Isak-Bego (commented):
Hey Buba!
Don't give up! You are almost there! Please take a look at my comments and see if you understand what my point is. Lets stay in touch and try to merge this ASAP. 
Thank you for your effort!

services/py-api/src/database/repository/participants_repository.py
```diff
@@ -21,27 +21,36 @@ class ParticipantsRepository(CRUDRepository):
     def __init__(self, db_manager: DatabaseManager, collection_name: str) -> None:
         self._collection = db_manager.get_collection(collection_name)
 
-    async def fetch_by_id(self, obj_id: str) -> Result:
-        return await self._collection.find_one({"_id": ObjectId(obj_id)})
+    async def fetch_by_id(self, obj_id: str) -> Result[Participant, ParticipantNotFoundError | Exception]:
+        try:
+            participant = await self._collection.find_one()
```
asynchroza, line 26: You have removed the arguments?

services/py-api/src/database/repository/participants_repository.py
```diff
@@ -21,27 +21,36 @@ class ParticipantsRepository(CRUDRepository):
...
+            LOG.exception(f"Failed to fetch participant with id {obj_id} due to err {e}")
+            return Err(e)
 
     async def fetch_all(self) -> Result:
         raise NotImplementedError()
 
     async def update(
         self, obj_id: str, input_data: BaseModel, session: Optional[AsyncIOMotorClientSession] = None, **kwargs: Any
-    ) -> Result[Ok, Exception]:
+    ) -> Result[Participant, ParticipantNotFoundError | Exception]:
```
asynchroza, line 42: For the sake of documentation you may leave this type as such but in the future you should know that `ParticipantNotFoundError` is a subset of `Exception`, therefore following set theory, a union of `ParticipantNotFoundError` and `Exception` is `Exception`.

If you become a TypeScript developer, this will be exceptionally useful. Or if python becomes more pedantic.

services/py-api/src/database/repository/participants_repository.py
```diff
@@ -21,27 +21,36 @@ class ParticipantsRepository(CRUDRepository):
...
         try:
             LOG.debug(f"Updating participant with id {obj_id}")
             result = await self._collection.find_one_and_update(
-                {"_id": ObjectId(obj_id)}, {"$set": input_data}, session=session
+                {"_id": ObjectId(obj_id)}, {"$set": input_data}, projection={"_id": 0}, session=session
             )
             if result:
                 LOG.debug(f"Successfully updated participant with id {obj_id}")
-                return Ok(result)
+                return Ok(Participant(id=obj_id, **result))
```
asynchroza, line 50: Your return type is not matching what you're returning. Update your return type.

services/py-api/src/database/repository/participants_repository.py
```diff
@@ -21,27 +21,36 @@ class ParticipantsRepository(CRUDRepository):
...
         self, obj_id: str, input_data: BaseModel, session: Optional[AsyncIOMotorClientSession] = None, **kwargs: Any
-    ) -> Result[Ok, Exception]:
+    ) -> Result[Participant, ParticipantNotFoundError | Exception]:
         try:
             LOG.debug(f"Updating participant with id {obj_id}")
             result = await self._collection.find_one_and_update(
-                {"_id": ObjectId(obj_id)}, {"$set": input_data}, session=session
+                {"_id": ObjectId(obj_id)}, {"$set": input_data}, projection={"_id": 0}, session=session
             )
             if result:
```
asynchroza, line 48: Update to: 

```python
if not result:
```

In order to avoid readability issues, you should keep the [happy path](https://en.wikipedia.org/wiki/Happy_path) on the left side.

asynchroza: Readability issues such as nesting and hard to follow error flow.

services/py-api/src/service/hackathon_service.py
```diff
@@ -70,27 +71,52 @@ async def create_random_participant(
...
+                return result_admin
+
+            result_team = await self._team_repo.update(
+                obj_id=team_id, input_data={"is_verified": True}, session=session
+            )
+            if is_err(result_team):
+                return result_team
+
+            return Ok((result_admin, result_team))
+        # verify random participant
```
asynchroza, line 119: This function is missing a return value for the default path (when not admin)

services/py-api/src/service/hackathon_service.py
```diff
@@ -70,6 +71,48 @@ async def create_random_participant(
...
+
+        # Updates the random participant if it exists
+        result = await self._participant_repo.update(jwt_data["sub"], {"email_verified": True})
+
+        if is_err(result):
+            return result
+
+        # As when first created, the random participant is not assigned to a team we return the team as None
+        return Ok((result.ok_value, None))
+
```
Isak-Bego, line 86: Remove this logic as it will be merged with Alexandr's pull request.

services/py-api/src/service/hackathon_service.py
```diff
@@ -70,6 +71,48 @@ async def create_random_participant(
...
+    async def verify_admin_participant_and_team_in_transaction(self, jwt_data: JwtUserData) -> Result[
+        Tuple[Participant, Team],
+        ParticipantNotFoundError | TeamNotFoundError | Exception,
+    ]:
+        return await self._tx_manager.with_transaction(self.verify_admin_participant_and_team_callback, jwt_data)
+
+    async def verify_admin_participant_and_team_callback(
```
Isak-Bego, line 93: Make this a private method by changing the name of the method to `_verify_admin_participant_and_team_callback()`

services/py-api/src/service/participants_verification_service.py
```diff
@@ -0,0 +1,35 @@
...
+    async def verify_random_participant(self, jwt_data: JwtUserData) -> Result[
+        Participant,
+        HackathonCapacityExceededError | Exception,
+    ]:
+        # Capacity Check 1
+        has_capacity = await self._hackathon_service.check_capacity_register_random_participant_case()
+        if not has_capacity:
+            return Err(HackathonCapacityExceededError())
+
+        return await self._hackathon_service.verify_random_participant(jwt_data=jwt_data)
```
Isak-Bego, line 35: Again this will be provided by Alexandr. Please delete it.

services/py-api/src/server/handlers/verification_handler.py
```diff
@@ -0,0 +1,49 @@
...
+            response.status_code = status.HTTP_400_BAD_REQUEST
+            return ErrResponse(error=result.err_value)
+
+        jwt_payload = result.ok_value
+
+        if jwt_payload["is_admin"]:
+            # Call for the admin participant case
+            result = await self._service.verify_admin_participant(jwt_data=jwt_payload)
+        else:
+            result = await self._service.verify_random_participant(jwt_data=jwt_payload)
```
Isak-Bego, line 30: Change this line with `result = Err(NotImplementedError())` as it will be provided by Alexandr in his PR

Isak-Bego (commented):
Just a few nits to address. The logic works as intended! I am so happy that you were able to pull all of this!! Good job!

The next step is to write unit tests for the individual functions and integration tests to test the workflow. Follow the examples that we have set. Let me know if you have any questions on the comments or tests

services/py-api/src/database/repository/participants_repository.py
```diff
@@ -17,20 +17,58 @@
...
+    ) -> Result[Participant, ParticipantNotFoundError | Exception]:
+        try:
+
+            LOG.debug(f"Updating participant with ObjectId={obj_id}, by setting {updated_data}.")
+
+            # ReturnDocument.AFTER returns the updated document instead of the orignal document which is the
+            # default behaviour.
+            result = await self._collection.find_one_and_update(
+                filter={"_id": ObjectId(obj_id)},
+                update={"$set": updated_data},
```
cl3vy, line 57: Shouldn't this method expect a Participant as updated_data, converting it then to a Dict in order to update should be handled by it.

services/py-api/tests/integration_tests/test_verification_routes.py
```diff
@@ -0,0 +1,118 @@
...
+async def test_verify_participant_admin_case_when_participant_is_not_found(
+    generate_participant_request_body: Callable[..., Dict[str, Any]],
+    create_test_participant: Callable[..., Dict[str, Any]],
+    mock_obj_id: str,
+    async_client: AsyncClient,
+) -> None:
+
+    # we need an existing team_id for the token and the only way to create a team is to create an admin participant as well
+    admin_participant_body = generate_participant_request_body(is_admin=True, team_name="test")
+    res_one = await create_test_participant(participant_body=admin_participant_body)
```
Isak-Bego, line 55: Creating a participant when you are testing the `participant not found case` seems redundant. Let's use the mock_obj_id as a sub. It should do the job.

services/py-api/tests/integration_tests/test_verification_routes.py
```diff
@@ -0,0 +1,118 @@
...
+@pytest.mark.asyncio
+async def test_verify_participant_admin_case_when_team_is_not_found(
+    generate_participant_request_body: Callable[..., Dict[str, Any]],
+    create_test_participant: Callable[..., Dict[str, Any]],
+    mock_obj_id: str,
+    async_client: AsyncClient,
+) -> None:
+
+    # we need an existing participant_id for the token so we generate an admin participant
+    admin_participant_body = generate_participant_request_body(is_admin=True, team_name="test")
```
Isak-Bego, line 75: Let's apply the same logic here. If we are testing the `team is not found` case then we probably don't need to create an admin participant which creates a team. Let's a random participant and keep the same logic below.

services/py-api/tests/integration_tests/test_verification_routes.py
```diff
@@ -0,0 +1,118 @@
+import pytest
+from typing import Any, Callable, Dict
+from httpx import AsyncClient
+
+from src.utils import JwtUtility
+from src.server.schemas.jwt_schemas.jwt_user_data_schema import JwtUserData
```
Isak-Bego, line 6: As I have mentioned in one of the comments of this PR, let's change the JwtUserData to: 
```Python
class JwtUserData(BaseTypedDict):
    sub: str
    is_admin: bool
    team_name: str
    team_id: str
    exp: float
    is_invite: bool
    exp: int
```

So that we don't have to rewrite the tests when that is changed

services/py-api/tests/integration_tests/test_verification_routes.py
```diff
@@ -0,0 +1,118 @@
...
+
+
+@pytest.mark.asyncio
+async def test_verify_participant_admin_case_invalid_token(
+    async_client: AsyncClient,
+) -> None:
+
+    fake_token_result = await async_client.patch(url=f"{PARTICIPANT_ENDPOINT_URL}/verify?jwt_token=FAKE_TOKEN")
+
+    assert fake_token_result.status_code == 400
```
Isak-Bego, line 118: I would say we don't need further logic for testing the validity of the token. We can keep the expired jwt token case, but the rest is tested by the integration tests we have written for the JwtUtility, so writing this would be redundant

services/py-api/tests/unit_tests/conftest.py
```diff
@@ -120,6 +121,17 @@ def participant_registration_service_mock() -> Mock:
...
+    """This is a mock obj of ParticipantVerificationService. To change the return values of its methods use:
+    `p_reg_service.method_name.return_value=some_return_value`"""
+
+    p_ver_service = Mock(spec=ParticipantVerificationService)
+    p_ver_service.verify_admin_participant.return_value = AsyncMock()
+
+    return p_ver_service
+
+
 @pytest.fixture
```
Isak-Bego, line 135: Nice job! Bravo!

services/py-api/tests/unit_tests/test_database_layer/test_participant_repository.py
```diff
@@ -125,3 +125,54 @@ async def test_delete_participant_general_exception(
...
+async def test_update_participant_general_exception(
+    db_manager_mock: Mock, mock_obj_id: str, repo: ParticipantsRepository
+) -> None:
+    db_manager_mock.get_collection.return_value.find_one_and_update = AsyncMock(side_effect=Exception("Error"))
+
+    result = await repo.update(mock_obj_id, {"email_verified": True})
+
+    assert isinstance(result, Err)
+    assert isinstance(result.err_value, Exception)
+    assert str(result.err_value) == "Error"
```
Isak-Bego, line 178: We don't need to include this check. Let's change it in every occurance

Isak-Bego (commented):
Overall nice job! There are somethings that need to be changed, however I am happy that you have gotten more comfortable with testing. I have not rigorously checked the unit tests but they should be fine. I trust you!

services/py-api/src/database/repository/participants_repository.py
Isak-Bego: Thank you for your feedback @asynchroza ❤️

services/py-api/src/service/hackathon_service.py
```diff
@@ -161,21 +164,58 @@ async def check_team_capacity(self, team_id: str) -> bool:
...
+    ]:
+
+        result_verified_admin = await self._participant_repo.update(
+            obj_id=jwt_data["sub"], obj_fields=UpdatedParticipant(email_verified=True), session=session
+        )
+
+        if is_err(result_verified_admin):
+            return result_verified_admin
+
+        result_verified_admin_dump_json = result_verified_admin.ok_value.dump_as_json()
```
IvanObreshkov, line 208: wtf

IvanObreshkov (approved):

PR #821, opened 2024-12-21
#820 Rework front end tests to run as a GitHub action

Aleks1908:


Isak-Bego (changes requested):
If it works it works. Good Job Alieks! Just one last thing. Add the label for the tests passing in the main README section similar to 
image

Isak-Bego (approved):
LGTM! Thank you Aliex!

PR #825, opened 2024-12-25
#824 Write Unit and Integration Tests for Delete Endpoints

### LINKED ISSUE #824: Write Unit Tests for the Delete Endpoints
## Brief description:
Changes that were introduced in [#813](https://github.com/AUBGTheHUB/monolith/pull/813) involved the creation of delete endpoints for both the participants and teams. Thus, there is a need to write unit tests for handler, service and repository methods, but also integration tests that test the deletion workflow.

Isak-Bego:
#824 Write Unit and Integration Tests for Delete Endpoints

## Added tests for the delete endpoints

resolves #824

### DIFF
--- a/services/py-api/src/server/handlers/hackathon_handlers.py
+++ b/services/py-api/src/server/handlers/hackathon_handlers.py
@@ -22,8 +22,9 @@ class HackathonManagementHandlers:
                 return ErrResponse(error="Could not find the team with the specified id")
 
             response.status_code = status.HTTP_500_INTERNAL_SERVER_ERROR
-            return ErrResponse(error="An unexpected error occurred during the deletion of Participant")
+            return ErrResponse(error="An unexpected error occurred during the deletion of Team")
 
+        response.status_code = status.HTTP_200_OK
         return TeamDeletedResponse(team=result.ok_value)
 
     async def delete_participant(
@@ -40,4 +41,5 @@ class HackathonManagementHandlers:
             response.status_code = status.HTTP_500_INTERNAL_SERVER_ERROR
             return ErrResponse(error="An unexpected error occurred during the deletion of Participant")
 
+        response.status_code = status.HTTP_200_OK
         return ParticipantDeletedResponse(participant=result.ok_value)
--- a/services/py-api/src/server/handlers/participants_handlers.py
+++ b/services/py-api/src/server/handlers/participants_handlers.py
@@ -50,4 +50,5 @@ class ParticipantHandlers:
             response.status_code = status.HTTP_500_INTERNAL_SERVER_ERROR
             return ErrResponse(error="An unexpected error occurred during the creation of Participant")
 
+        response.status_code = status.HTTP_201_CREATED
         return ParticipantRegisteredResponse(participant=result.ok_value[0], team=result.ok_value[1])
--- a/services/py-api/tests/integration_tests/conftest.py
+++ b/services/py-api/tests/integration_tests/conftest.py
@@ -82,3 +82,24 @@ async def create_test_participant(async_client: AsyncClient) -> Callable[..., Di
         if result.status_code == 201:
             LOG.debug("Cleaning up test participant")
             await clean_up_test_participant(async_client=async_client, result_json=result.json())
+
+
+@pytest_asyncio.fixture
+def generate_participant_request_body() -> Callable[..., Dict[str, Any]]:
+    def _generate_participant_request_body(
+        name: str | None = "testtest",
+        email: str | None = "testtest@test.com",
+        is_admin: bool | None = False,
+        **kwargs: Any,
+    ) -> Dict[str, Any]:
+
+        request_body = {"name": name, "email": email, "is_admin": is_admin, **kwargs}
+        # A new dict without None values
+        return {key: value for key, value in request_body.items() if value is not None}
+
+    return _generate_participant_request_body
+
+
+@pytest.fixture
+def mock_obj_id() -> str:
+    return "507f1f77bcf86cd799439011"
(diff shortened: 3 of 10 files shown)


services/py-api/src/server/handlers/participants_handlers.py
```diff
@@ -50,4 +50,5 @@ async def create_participant(
             response.status_code = status.HTTP_500_INTERNAL_SERVER_ERROR
             return ErrResponse(error="An unexpected error occurred during the creation of Participant")
```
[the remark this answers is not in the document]
asynchroza: I agree

services/py-api/tests/integration_tests/conftest.py
```diff
@@ -82,3 +82,24 @@ async def _create(participant_body: Dict[str, Any]) -> Response:
...
+        email: str | None = "testtest@test.com",
+        is_admin: bool | None = False,
+        **kwargs: Any,
+    ) -> Dict[str, Any]:
+
+        request_body = {"name": name, "email": email, "is_admin": is_admin, **kwargs}
+        # A new dict without None values
+        return {key: value for key, value in request_body.items() if value is not None}
+
+    return _generate_participant_request_body
```
asynchroza, line 100: I think it would make more sense to call the function `get_participant_request_body_generator` or something of the sorts

```python
generate_participant_request_body = get_participant_request_body_generator()
request_body = generate_participant_request_body("Misho", "notmyemail@gmail.com", True)
```

services/py-api/tests/integration_tests/test_participant_routes.py
```diff
@@ -142,3 +150,62 @@ async def test_create_admin_participant_email_already_exists(
     resp_json = resp.json()
 
     assert resp_json["error"] == "Participant with this email already exists"
+
+
+@patch.dict(environ, {"SECRET_AUTH_TOKEN": "OFFLINE_TOKEN"})
```
asynchroza, line 155: Are all db connections mocked? This is a security vulnerability otherwise.

services/py-api/src/server/handlers/hackathon_handlers.py
```diff
@@ -40,4 +41,5 @@ async def delete_participant(
             response.status_code = status.HTTP_500_INTERNAL_SERVER_ERROR
             return ErrResponse(error="An unexpected error occurred during the deletion of Participant")
 
+        response.status_code = status.HTTP_200_OK
```
asynchroza, line 44: Why is the delete handler in hackathon but the create one is in participants?

services/py-api/tests/integration_tests/test_participant_routes.py
asynchroza: Never mind, I'm sleeping. That's how it's supposed to be. Thank you.

services/py-api/tests/integration_tests/conftest.py
Isak-Bego: I have used the same pattern as it is noted in the documentation of pytest fixtures. Here is a link to the docs: https://docs.pytest.org/en/6.2.x/fixture.html#factories-as-fixtures

services/py-api/src/server/handlers/hackathon_handlers.py
Isak-Bego: Alright, so since the Hackathon Service is the only service that direct access to the repositories as for the current structure, we were kinda forced to write the delete logic there, because every other implementation would either break encapsulation or wasn't what we were looking for in terms of structure. This way, we decided to keep it for all the hackathon management business logic (which would be utilized by an admin user), while the creation of a participant is mainly utilized by the clients (people who want to register). We are aware that this might not be ideal, its one of the few solutions that we came up with. 
I am open to suggestions if you have any. Thank you Misho!

services/py-api/tests/integration_tests/conftest.py
asynchroza: Functionality is correct. I'm not making a comment on this but rather the naming of the fixture.

services/py-api/src/server/handlers/hackathon_handlers.py
asynchroza: Why do you have the participant service then?

services/py-api/tests/integration_tests/conftest.py
Isak-Bego: Oh I see that's right.

services/py-api/src/server/handlers/hackathon_handlers.py
Isak-Bego: In the participant service we have the capacity checks and JWT decoding and checks (in the case of link participant), which in the case that are not successful, return a response without the need of going deeper into the flow. 
The Hackathon Service on the other hand has the business logic for the CRUD operations (more checks).
Again, this is far from ideal and I will reconsider refactoring it in the future. Ivan created the Hackathon Service to distribute the logic of Participant Service which was turning into an iceberg class as we ended up mocking private methods.

services/py-api/tests/integration_tests/conftest.py
Isak-Bego: Fixed in 48e6dac35f8a640f6d1f74cd7330b33974f613fd

Isak-Bego:
@asynchroza Do you approve this one?

services/py-api/src/server/handlers/hackathon_handlers.py
asynchroza: Yeah. This is not thought out well.

asynchroza:
> @asynchroza Do you approve this one?

As long as it works 👍

asynchroza (approved):

PR #826, opened 2024-12-27
#805 register link participant

### LINKED ISSUE #805: feature: implement the registration of participant with registration link
## Brief description:

STORY:
>"Imagine a group of friends that want to register to our hackathon. One of the friends registers himself as a team admin by creating the team with the name that you all decided. He verifies himself and he gets an email from the hub that contains a link that he can share with the rest of the team. He shares the link with the rest of his friends. They click on it and now they can register in the team that was created by the first friend. "

1. We should provide all the logic that makes this possible. We get a request whose body contains the information of the participant that wants to join the team, and that has a [query parameter](https://www.branch.io/glossary/query-parameters/) with the JWT token that was generated by us and passed to the link the admin received with his email. 

2. Further more we send the request body and JWT token form the route function down to the handler layer where the correct course of action is decided. 

3. Is this an admin who is trying to register? Is this a random participant? Or maybe is it a participant with a link? A participant with a link has a team_name on the body of the request that is provided by the frontend and has the is_admin field as false. Once we decide this we get down to the service layer where the rest of the logic is going to be performed (business logic).

Here is a link to the API design schema: https://excalidraw.com/#json=qJbg3VyVnKP_jOAlDq2Kf,BileoGllBKsbeQOf8KydlQ
You should implement Flow 2 (in Blue): 

image

## How to achieve it:
1. Start by adding the query parameter jwt_token to the create participant (POST) route. It should be Optional since not all the requests to this route will have the jwt_token as a query parameter and the type either str or None. See more about how you can acheive this here: https://fastapi.tiangolo.com/tutorial/query-params-str-validations/

2. Pass the jwt_token to the handler method that is being called within this function

3. In the handler method write the condition that determines that this request belongs to a participant with a link. (with a team_name, is_admin=false). Add it somewhere here: 

Filepath: `src/server/handlers/participant_handlers.py`

```Python

class ParticipantHandlers:
    def __init__(self, service: ParticipantRegistrationService) -> None:
        self._service = service

    async def create_participant(
        self, response: Response, input_data: ParticipantRequestBody
    ) -> ParticipantRegisteredInTeamResponse | ErrResponse:
        # TODO:
        # When the logic for all cases is done
        # if input_data.is_admin and input_data.team_name:
        #     result = await self._service.register_admin_participant(input_data)
        # else:
        #     ...
        result = await self._service.register_admin_participant(input_data)
...
```

4. Create a register_invite_link_participant() method in the ParticipantsService where you will write the buisness logic. To acheive this follow the flowchart that I have attached above. 

5. Before you check if the team has enough free capacity to accomodate the participant please check if the team with team_id from the jwt token has the same name as the team_name provided in the body of the request (input_data). Then proceed with the check. 

Please stay strict to the Repository-Service-Handler pattern to acheive it. 
More about it here: https://tom-collings.medium.com/controller-service-repository-16e29a4684e5

**The ticket will be considered complete when unit and integration tests are also written for it**

Please ask questions.

KDimitrov333:
# #805 Implement Registration of Participant with Registration Link

## Added functionality for registration of non-admin participants through a provided link:
Now, after an admin for a team has been registered and a team created, they can send the invite link to their teammates, through which they can register. Unit tests remain to be written. Also verification with emails as I am not sure if we fixed the problems there and if we can do it now.

resolves #805   

## How to verify:

- [ ] Start the Python Backend

- [ ] Open your browser or API client and head to (https://localhost:8080/api/v3/hackathon/participants?jwt_token=

- [ ] Use a valid jwt_token

- [ ] Read the server's message - created (201) or an error (409, 500)?


services/py-api/src/database/repository/teams_repository.py
```diff
@@ -96,3 +96,15 @@ async def get_verified_registered_teams_count(self) -> int:
         # Ignoring mypy type due to mypy err: 'Returning Any from function declared to return "int"  [no-any-return]'
         # which is not true
         return await self._collection.count_documents({"is_verified": True})  # type: ignore
+    
+    async def fetch_team_by_name(self, team_name: str) -> Result:
```
Isak-Bego, line 100: change the name to fetch_by_team_name to match the rest of fetch functions

services/py-api/src/database/repository/teams_repository.py
```diff
@@ -96,3 +96,15 @@ async def get_verified_registered_teams_count(self) -> int:
...
+        """Fetches a team by the team_name from the participant"""
+        try:
+            team = await self._collection.find_one({"name": team_name})
+            
+            if team is None:  # If no team is found, return an Err
+                return Err(Exception(f"Team with name '{team_name}' not found"))
```
Isak-Bego, line 106: There is a custom exception for this purpose. You can find it in exceptions.py. 
So instead of this you can return Err(TeamNotFoundError()).

services/py-api/src/database/repository/teams_repository.py
```diff
@@ -96,3 +96,15 @@ async def get_verified_registered_teams_count(self) -> int:
...
+
+            return Ok(team)
+        except Exception as e:
+            return Err(e)
```
Isak-Bego, line 110: Use a similar pattern when returning the team and the general exception: 
```Python
async def delete(
        self, obj_id: str, session: Optional[AsyncIOMotorClientSession] = None
    ) -> Result[Team, TeamNotFoundError | Exception]:
        """
        Deletes the team which corresponds to the provided object_id
        """
        try:
            LOG.debug("Deleting team...", team_obj_id=obj_id)
            """
            According to mongodb docs result is of type _DocumentType:
            https://pymongo.readthedocs.io/en/4.8.0/api/pymongo/collection.html#pymongo.collection.Collection.find_one_and_delete
            _id is projected because ObjectID is not serializable.
            We use the Team data class to represent the deleted participant.
            """
            result = await self._collection.find_one_and_delete(filter={"_id": ObjectId(obj_id)}, projection={"_id": 0})
            """
            The result is None when the team with the specified ObjectId is not found
            """
            if result:
                return Ok(Team(id=obj_id, **result))
            return Err(TeamNotFoundError())
        except Exception as e:
            LOG.exception("Team deletion failed due to err {}".format(e))
            return Err(e)
```

services/py-api/src/database/repository/teams_repository.py
Isak-Bego, line 100: Update the return type from Result to specifying the type of a Ok result and the type of an Err result. The format is 
`Result[OkType, ErrType]`

For example: 
```Python
async def delete(
        self, obj_id: str, session: Optional[AsyncIOMotorClientSession] = None
    ) -> Result[Team, TeamNotFoundError | Exception]:
```

services/py-api/src/server/handlers/participants_handlers.py
```diff
@@ -21,14 +25,27 @@ def __init__(self, service: ParticipantRegistrationService) -> None:
...
+            if is_err(decoded_result):
+                response.status_code = status.HTTP_401_UNAUTHORIZED
+                return ErrResponse(error=f"JWT decoding failed: {decoded_result.err_value}")
+            
+            decoded_data = decoded_result.ok_value
+            team_name_from_token = decoded_data["team_id"]
+
+            if input_data.team_name != team_name_from_token:
+                response.status_code = status.HTTP_400_BAD_REQUEST
+                return ErrResponse(error="Team name in request does not match the invite link.")
```
Isak-Bego, line 46: Move the token validation logic inside of the service layer. If you have any issues with decoding the token catch the exceptions in the handler. Remember we use the services to write all the business logic.

Only 
```Python
result = await self._service.register_invite_link_participant(input_data)
``` 

Should stay in the handler

services/py-api/src/server/handlers/participants_handlers.py
```diff
@@ -46,6 +63,10 @@ async def create_participant(
             if isinstance(result.err_value, HackathonCapacityExceededError):
                 response.status_code = status.HTTP_409_CONFLICT
                 return ErrResponse(error="Max hackathon capacity has been reached")
+            
+            if isinstance(result.err_value, TeamCapacityExceededError):
+                response.status_code = status.HTTP_409_CONFLICT
+                return ErrResponse(error="Your team's capacity has been reached")
```
Isak-Bego, line 69: Change the comment to "Max team capacity has been reached to match the same structure as the error above.

services/py-api/src/server/routes/participants_routes.py
```diff
@@ -39,9 +40,9 @@ def _h_handler(
     "", status_code=201, responses={201: {"model": ParticipantRegisteredResponse}, 409: {"model": ErrResponse}}
 )
 async def create_participant(
-    response: Response, input_data: ParticipantRequestBody, handler: ParticipantHandlers = Depends(_p_handler)
+    response: Response, input_data: ParticipantRequestBody, jwt_token: Optional[str]=Query(None), handler: ParticipantHandlers = Depends(_p_handler)
```
Isak-Bego, line 43: This works, however, I would prefer if we followed the fastapi documentation for defining optional query parameters
Here is a link to the docs: https://fastapi.tiangolo.com/tutorial/query-params/#optional-parameters
Here is an example from the docs: 
```Python
from typing import Union

from fastapi import FastAPI

app = FastAPI()


@app.get("/items/{item_id}")
async def read_item(item_id: str, q: Union[str, None] = None):
    if q:
        return {"item_id": item_id, "q": q}
    return {"item_id": item_id}
```

where q is the query parameter

services/py-api/src/server/handlers/participants_handlers.py
```diff
@@ -21,14 +25,27 @@ def __init__(self, service: ParticipantRegistrationService) -> None:
...
-        #     result = await self._service.register_invite_participant(input_data)
+        elif input_data.is_admin is False and input_data.team_name:
+            # Decode the token
+            decoded_result = JwtUtility.decode_data(token=jwt_token, schema=JwtUserData)
+            if is_err(decoded_result):
+                response.status_code = status.HTTP_401_UNAUTHORIZED
+                return ErrResponse(error=f"JWT decoding failed: {decoded_result.err_value}")
+            
+            decoded_data = decoded_result.ok_value
+            team_name_from_token = decoded_data["team_id"]
```
Isak-Bego, line 42: team_name is not the same as team_id. The the name property of a team is a string while the _id property of a team is an Object Id, so you will need to change this check.

services/py-api/src/server/handlers/participants_handlers.py
```diff
@@ -21,14 +25,27 @@ def __init__(self, service: ParticipantRegistrationService) -> None:
...
 
-        # TODO: Implement this when the invite participant case is done
-        # elif input_data.is_admin is False and input_data.team_name:
-        #     result = await self._service.register_invite_participant(input_data)
+        elif input_data.is_admin is False and input_data.team_name:
+            # Decode the token
+            decoded_result = JwtUtility.decode_data(token=jwt_token, schema=JwtUserData)
+            if is_err(decoded_result):
+                response.status_code = status.HTTP_401_UNAUTHORIZED
+                return ErrResponse(error=f"JWT decoding failed: {decoded_result.err_value}")
```
Isak-Bego, line 39: you can just print error=decoded_result.err_value

decoded_result.err_value could be: 
1. "The decoded token does not correspond with the provided schema."
2. "The JWT token has expired."
3. "The JWT token has invalid signature."

So these messages are sufficient on their own

services/py-api/src/server/handlers/participants_handlers.py
```diff
@@ -1,11 +1,15 @@
+from typing import Optional
 from result import is_err
+from src.server.schemas.jwt_schemas.jwt_user_data_schema import JwtUserData
```
Isak-Bego, line 3: Change JwtUserData to have: 
the following fields: 
```Python
# All the types within this should be required
class JwtUserData(BaseTypedDict):
    sub: str
    is_admin: bool
    team_name: str
    is_invite: bool
    exp: datetime
```

services/py-api/src/service/hackathon_service.py
```diff
@@ -69,6 +69,22 @@ async def create_random_participant(
...
+        if is_err(participant_result):
+            return participant_result
+
+        # Assign the team whose name matches that from input_data.team_name
+        team_result = await self._team_repo.fetch_team_by_name(input_data.team_name)
+        if is_err(team_result):
+            return team_result
+        
+        # Return the new participant and the preexisting team
+        return Ok((participant_result.ok_value, team_result.ok_value))
```
Isak-Bego, line 87: We only return the participant

services/py-api/src/service/hackathon_service.py
```diff
@@ -69,6 +69,22 @@ async def create_random_participant(
...
+    ) -> Result[Tuple[Participant, Team], DuplicateEmailError | Exception]:
+        
+        participant_result = await self._participant_repo.create(input_data)
+        if is_err(participant_result):
+            return participant_result
+
+        # Assign the team whose name matches that from input_data.team_name
+        team_result = await self._team_repo.fetch_team_by_name(input_data.team_name)
+        if is_err(team_result):
+            return team_result
```
Isak-Bego, line 84: I dont think we need to return the team too, since we are not adding a new team or updating a new one, so you can omit this.

services/py-api/src/service/hackathon_service.py
```diff
@@ -69,6 +69,22 @@ async def create_random_participant(
...
         # As when first created, the random participant is not assigned to a team we return the team as None
         return Ok((result.ok_value, None))
+    
+    async def create_invite_link_participant(
+            self, input_data: ParticipantRequestBody
+    ) -> Result[Tuple[Participant, Team], DuplicateEmailError | Exception]:
+        
+        participant_result = await self._participant_repo.create(input_data)
+        if is_err(participant_result):
+            return participant_result
```
Isak-Bego, line 79: Make sure that this newly created participant has the field is_verified=True (participants with link are automatically verified and in the team_id field that have the id of the team with the same name as provided in the jwt token). In other words make sure they are present in the input data.

services/py-api/src/service/participants_registration_service.py
```diff
@@ -42,3 +43,15 @@ async def register_random_participant(self, input_data: ParticipantRequestBody)
...
         return await self._hackathon_service.create_random_participant(input_data)
+    
+    async def register_invite_link_participant(self, input_data: ParticipantRequestBody) -> Result[
+        Tuple[Participant, Team],
+        DuplicateEmailError | DuplicateTeamNameError | HackathonCapacityExceededError | Exception,
+    ]:
+        # Check Team Capacity
+        has_capacity = await self._hackathon_service.check_team_capacity(input_data.team_name)
+        if not has_capacity:
+            return Err(TeamCapacityExceededError())
```
Isak-Bego, line 54: Suggestion: you can do the decoding of the jwt here and move the rest of the logic in the hackathon service.

services/py-api/src/database/repository/teams_repository.py
KDimitrov333: Fixed in commit a9a6b60a1c533cba63bd89fc5bee4362218c00f7

services/py-api/src/database/repository/teams_repository.py
KDimitrov333: Fixed in commit a9a6b60a1c533cba63bd89fc5bee4362218c00f7

services/py-api/src/database/repository/teams_repository.py
KDimitrov333: Fixed in commit a9a6b60a1c533cba63bd89fc5bee4362218c00f7

services/py-api/src/database/repository/teams_repository.py
KDimitrov333: Fixed in commit a9a6b60a1c533cba63bd89fc5bee4362218c00f7

services/py-api/src/server/handlers/participants_handlers.py
KDimitrov333: Fixed in commit a9a6b60a1c533cba63bd89fc5bee4362218c00f7

services/py-api/src/server/handlers/participants_handlers.py
KDimitrov333: Fixed in commit a9a6b60a1c533cba63bd89fc5bee4362218c00f7

services/py-api/src/server/routes/participants_routes.py
KDimitrov333: Fixed in commit a9a6b60a1c533cba63bd89fc5bee4362218c00f7

services/py-api/src/server/handlers/participants_handlers.py
KDimitrov333: Ok but since I moved this logic to the service layer I should return the Err not an Error Response right? Then the ErrorResponse should be in the handler? If so should I create new exceptions for these messages so we can use the same isinstance pattern?

services/py-api/src/server/handlers/participants_handlers.py
KDimitrov333: Fixed in commit a9a6b60a1c533cba63bd89fc5bee4362218c00f7

services/py-api/src/service/hackathon_service.py
KDimitrov333: Fixed in commit a9a6b60a1c533cba63bd89fc5bee4362218c00f7

services/py-api/src/service/hackathon_service.py
KDimitrov333: I will use it to get the team id but tell me if there is still a way to get rid of it

services/py-api/src/service/hackathon_service.py
KDimitrov333: Fixed in commit d8f0832f77647d818b784c5ad0942048bd62a834

services/py-api/src/service/participants_registration_service.py
KDimitrov333: Fixed in commit d8f0832f77647d818b784c5ad0942048bd62a834

KDimitrov333:
I think the integration tests aren't going through because I had to change the JwtUserData. What should I do?

Isak-Bego (commented):
Hey Kalo, 
Good job on the PR. There are some small nits that should be changed. I will push a patch on your branch later today and all that will be left for you is to write unit and integration tests. 
Good job!
A small note: If you ever need any help with testing the application logic just text me and I will be willing to help you.

Isak-Bego:
> I think the integration tests aren't going through because I had to change the JwtUserData. What should I do?

I will fix the integration test regarding the jwt functionality for you

IvanObreshkov:
Closing as the work was finished in https://github.com/AUBGTheHUB/monolith/pull/839, Thank you @KDimitrov333 you provided a good base.

PR #849, opened 2025-01-25
835 implement endpoint for sending verification emails

### LINKED ISSUE #835: feature: implement endpoint for sending verification emails
## Brief description:
We aim to provide our users with the functionality of being able to retry sending a verification email if they don't receive it. This way the process of verifying is safer. 

For this reason, we have to have an endpoint that the client side can trigger by passing the email or participant object id that will retry sending a verification email. This endpoint can be an extension of the already established verification endpoint `/api/v3/hackathon/participants/verify`. A suggestion could be `/api/v3/hackathon/participants/verify/send-email`

## How to achieve it:
We should rate limit this process in case someone tries to spam the endpoint and rate limit our resend service (since with the free tier we can only send so many emails within a given time). @cl3vy has already provided a working solution. We firstly need to update the structure of the `participant` document to: 

```python
@dataclass(kw_only=True)
class Participant(Base):
    """A representation of the Participant entity in Mongo. It is also the schema of how the entity should look
    like in Mongo before it is inserted"""
    name: str
    email: EmailStr
    is_admin: bool
    email_verified: bool = field(default=False)
    team_id: Optional[SerializableObjectId]
    last_sent_email: Optional[datetime] = field(default=None)
```

Then every time we get a request in this endpoint, we check if enough time has passed since the last email sent. 

```python
        # Check rate limit
        if participant.last_sent_email:
            if datetime.now() - participant.last_sent_email < timedelta(seconds=RATE_LIMIT_SECONDS):
                LOG.error(f"Rate limit exceeded for participant {participant.id} with email {participant.email}.")
                return Err(EmailRateLimitExceededError())
```

If the rate limit is not exceeded you proceed with sending the email. 
You can get some inspiration from these changes: 600e52ddbd0ace985410ea9f0358ca0d7c0e799b

> The issue will be considered done when the desired functionality is achieved and there are unit tests for each method that you implement.

## Prerequisites
The following issues should be closed before starting work with this one:
- [x] #807
- [x] #808 
- [x] #744

Bubomira:
#835 feature: implement endpoint for sending verification emails

1)created an endpoint for resending verification emails 
2)created the send_verification_email in the verification_handler
3)created the send_verification_email in the verification_service
4)created the check_send_verification_email_rate_limit() in hackaton service, which firstly checks if the particpant exists and if it does then checks if the rate limit is satisfyed and the send_verification_email, which is going to send the email (part of another issue) and update the last_sent_email field
5)added the ParticipantAlreadyVerifiedError and the EmailRateLimitExceededError
6)added VerificationEmailSentSuccessfullyResponse
7)added last_sent_email field to Participant entity
8) unit_tests

resolves #835


services/py-api/src/service/hackathon_service.py
```diff
@@ -223,3 +226,40 @@ async def delete_participant(
...
+
+        is_within_rate_limit = datetime.now() - participant.ok_value.last_sent_email >= timedelta(
+            seconds=self.RATE_LIMIT_SECONDS
+        )
+        return Ok(is_within_rate_limit)
+
+    async def send_verification_email(
+        self, participant_id: str
+    ) -> Result[Participant, ParticipantNotFoundError | Exception]:
+        # TODO: send email like a background task (part of another issue)
```
Isak-Bego, line 256: You can go ahead an use Klevi's Email Service that is now merged into main to send a verification email. Note that there is a specific function 
```Python
async def send_participant_verification_email(
        self,
        participant: Participant,
        team_name: str,
        confirmation_link: str,
    ) -> Result[Email, str]:
```

in the resend_mailing service that does that.

services/py-api/tests/unit_tests/test_handlers/test_verification_handlers.py
```diff
@@ -248,3 +259,71 @@ async def test_verify_random_hackathon_capacity_reached(
...
+
+    participant_verification_service_mock.send_verification_email.assert_awaited_once()
+
+    assert isinstance(result, Response)
+    assert isinstance(result.response_model, VerificationEmailSentSuccessfullyResponse)
+    assert result.status_code == status.HTTP_200_OK
+    assert result.response_model.participant.name == TEST_USER_NAME
+    assert result.response_model.participant.email == TEST_USER_EMAIL
+    assert result.response_model.participant.is_admin is True
+
```
Isak-Bego, line 282: These tests loook cleeeaaaan. I am super proud of you! Amazing, Amazing job!

services/py-api/tests/unit_tests/test_service_layer/test_hackathon_service.py
Isak-Bego, line 1: Nice ones!

Isak-Bego (commented):
Everything so far looks perfect. You just need to add the functionality for sending the actual verification email which is not provided. Keep going in this direction! I am impressed! Clean, Easy-to-understand and well-tested

PR #858, opened 2025-01-30
Secure the staging env using HTTP

Isak-Bego:
# Reference the ISSUE in the beggining of the title of the PR
e.g. #129 Added webhooks for triggering something (#130)

## Added webhooks for triggering something:
Flippity-floop webbywoos infused with ziggly-zaggly code snippets, now dance harmoniously with the techno-tunes of binary bongo beats! Behold, the magnificent symphony of cybernetic squirrels tap-dancing upon the digital drums of data, while quantum quokkas spin around in an ethereal waltz. These webhooks, those mystical connectors of the byte realm, have donned their pixelated tutus to trigger a cascade of cosmic confetti, launching rocket ships fueled by electrified spaghetti into the intergalactic spaghetti meatball of automation! 🚀🍝💻

resolves #129  

## How to verify:


services/web/nginx.dev.conf
```diff
@@ -12,6 +12,9 @@ server {
     # giving information about what we should expect to find in the request body
     include mime.types;
 
+    auth_basic "Restricted Access";
+    auth_basic_user_file /conf/basic-auth/nginx/.htpasswd;
```
IvanObreshkov, line 16: So the `.htpasswd` file is created and stored on the Dev machine at `/etc/nginx/.htpasswd`, right? The content of this file is the hashed password. If yes to both of these, merge.

IvanObreshkov (approved):

docker-swarm-stack.yml
```diff
@@ -22,6 +22,7 @@ services:
       - py-api
     volumes:
       - /etc/letsencrypt:/etc/letsencrypt
+      - /etc/nginx:/conf/basic-auth/nginx
```
IvanObreshkov, line 25: This will most probably fail on prod, as `/etc/nginx` does not exist. We should experiment with a named volumes managed by docker

IvanObreshkov (changes requested):

docker-swarm-stack.yml
Isak-Bego: Yes bro this is just a patch to protect dev. from getting indexed

PR #869, opened 2025-02-06
#860 feature switch

### LINKED ISSUE #860: feature: create a feature switch for the registration api
## Brief description:
Our API is meant to mainly serve for the registration of the hackathon. For this reason, the endpoints that provide the registration/verification functionality should not be accepting requests when the registration is closed. 
## How to achieve it:
We should create a function which will be added as a dependency to the registration routes following a Fastapi pattern: 
https://fastapi.tiangolo.com/reference/dependencies/#fastapi.Depends
We already have examples of this in our codebase. For example you can check the team_routes.py: 
```Python
from fastapi import APIRouter, Depends
from src.service.hackathon_service import HackathonService
from src.server.handlers.hackathon_handlers import HackathonManagementHandlers
from src.server.routes.dependency_factory import _h_service
from src.server.schemas.response_schemas.schemas import ErrResponse, TeamDeletedResponse
from starlette.responses import Response
from src.server.routes.dependency_factory import is_auth, validate_obj_id


teams_router = APIRouter(prefix="/hackathon/teams")


def _handler(
    h_service: HackathonService = Depends(_h_service),
) -> HackathonManagementHandlers:
    return HackathonManagementHandlers(h_service)


@teams_router.delete(
    "/{object_id}",
    status_code=200,
    responses={200: {"model": TeamDeletedResponse}, 404: {"model": ErrResponse}},
    dependencies=[Depends(is_auth), Depends(validate_obj_id)],
)
async def delete_team(object_id: str, handler: HackathonManagementHandlers = Depends(_handler)) -> Response:
    return await handler.delete_team(object_id)

```

We want something similar. that we can append to routes in a Depends().

In case this feature switch is not open then you return a response like: status_code: 409, "Registration is closed". Otherwise you should let the request be processed. 

You might also have to edit the repositories a little and create a repository layer for the feature switches.

1. Creating a feature switch pydantic model (check the participant and team models) that you are going to use for the collection named feature-switches in the DB. The structure should be simple: "feature": str,  "state": bool (true or false)

2. Create a Feature switch repository that implements the base repository methods. 

3. Create a Feature Switch Service that is instantiated with the Feature Switch repository. 

4. Create a Feature Switch Handler that is instantiated with the Feature Switch Service. 

5. We should not create a route for the feature switches as for now, as we don't have a form of authentication.

6. Create a Dependency function that calls a method in the Feature Switch Handler that checks the status of the isRegistrationOpen feature switch and provides the functionality that I have described in the beginning of this issue. 

Please ask me clarification questions if you have any. 

> The issue will be considered done when there are appropriate tests are written for the functionality you provide.

KDimitrov333:
#860 Create feature switch

resolves #860


services/py-api/src/database/repository/feature_switch_repository.py
```diff
@@ -0,0 +1,20 @@
+from typing import Optional
+from pymongo.collection import Collection
+from src.server.schemas.feature_switch_schemas.feature_switch_schema import FeatureSwitch
+
+class FeatureSwitchRepository:
```
Isak-Bego, line 5: This should inherit from CRUDRepository. Check the examples of the other repos.

Isak-Bego (commented):
This can serve as a good base. Can you create a collection inside our database using the MongoDB Compass and actually test the feature switch somehow? Let me know if you will be able to acheive this. In any case I can jump in and help.

IvanObreshkov (approved):

PR #870, opened 2025-02-06
send verification emails upon successful registration

Isak-Bego:
send verification emails upon successful registration

## Send a verification email once the participant is registered:
We should send a verification email as a background task once the participant is registered so that they can verify themselves. 

Note: This PR is still work in progress. It is not ready for review. It shall rather serve as a handy example on how you can use the mailing service and how to use fastapi's Background Tasks.


Aleks1908 (approved):
LGTM

Issue #874, opened 2025-02-08
Fix hover on landing and spacing between sections


DemonHashira:
The hover on the landing part is too big. For best usability, it should be smaller. 
For the margin limit it a bit because this is unnecessary space right now. 

![Image](https://github.com/user-attachments/assets/e855772d-6441-4006-97c4-bcf92fb1857e)

![Image](https://github.com/user-attachments/assets/9e60a83e-1a65-4687-bf05-56b7f7693c2a)

PR #884, opened 2025-02-09
Fixes 874: Make circuit effect smaller, make it disappear when leaving the section of the page and tweak the margins

Alexandr-Ignatov:
#874 
## Make Effect smaller and make it disappear

Made the circuit effect smaller when hovering the mouse. 
Also, now when the mouse leaves the blue section of the page the effect disappears (before it stayed in the place where mouse was the last time.)
Additionally, tweaked some margins.


DemonHashira (approved):
LGTM

PR #886, opened 2025-02-09
fixed issue with border on x

### LINKED ISSUE #873: Mobile Navigation
The problem is that when you open the menu on mobile, the close button is selected by default and when you click anywhere else in the menu it becomes normal. I can't upload a video here but once you go and check it you will see. 

Image

asiahaslam:
#873 fixed issue where x on mobile menu had a box around it by default

resolves #873


Aleks1908 (approved):
Nice job

Issue #871, opened 2025-02-07
Favicon and title of the pages


DemonHashira:
As can be seen from the picture we don't have a favicon for the site and the page's name should be another one.
This is a small fix that should take 10 minutes. You can use the favicon that we currently have on the site.

<img width="232" alt="Image" src="https://github.com/user-attachments/assets/09dbfd3c-ec38-42ee-bb48-b0055e607367" />

PR #889, opened 2025-02-09
871: Add favicon and title

Alexandr-Ignatov:
# #871


DemonHashira (changes requested):
Dont set the favicon like that. Go to master, find it there and include it in the project here.

DemonHashira (approved):
LGTM

PR #891, opened 2025-02-09
fix border and cyan blob

### LINKED ISSUE #868: Fix landing page issues on mobile devices
There seems to be an issue with the blob and the borders of the buttons on actual mobile devices
Image

asiahaslam:
#868 fixed issue with landing section blob/borders for safari on mobile

I tested it by making my browser window narrow on firefox, chrome, and safari and it only hides the cyan blob on safari (though the color is lighter on chrome than on firefox)

When the browser window is full size, it does not hide the blob

I also doubled the width of the button borders to hopefully resolve that other issue.

resolves #868


Aleks1908 (approved):
LGTM

PR #898, opened 2025-02-09
#895 hackaubg footer section

### LINKED ISSUE #895: HackAUBG Footer Section
We want to reuse the same navigation from the Main page in the HackAUBG page. Your jobs is to reuse it and rework the colors to fir the theme of the HackAUBG page which includes removing the blob

asiahaslam:
#895 added footer for the hackathon page
resolves #895


Aleks1908 (approved):
LGTM

PR #900, opened 2025-02-13
#835 implement endpoint for sending verification emails ivan rewrite

### LINKED ISSUE #835: feature: implement endpoint for sending verification emails

IvanObreshkov:
## What has changed:
* Refactored the existing MailService interface to be -> HackathonMailService(mail_client=MailCleint),
  * where in the HackathonMailService is the business logic of loading html templates for particular participants and offloading the sending of emails to BackgroundTasks
  * and the MailClient is a new interface which different Mail Clients should implement, to allow easy swapping between them. Currently we use only Resend client.

* Refactor hackathon service to use new mail service
* Fixed some type and error issues
* Fixed tests and added new ones.


services/py-api/src/service/hackathon_service.py
```diff
@@ -295,42 +332,77 @@ async def send_verification_email(
...
-        else:
+        if is_prod_env():
+            verification_link = f"https://{DOMAIN}{self._PARTICIPANTS_VERIFICATION_ROUTE}?jwt_token={jwt_token}"
+        elif is_dev_env():
             verification_link = (
-                f"https://dev.thehub-aubg.com/api/v3/hackathon/participants/verify?jwt_token={jwt_token}"
+                f"https://{SUBDOMAIN}.{DOMAIN}{self._PARTICIPANTS_VERIFICATION_ROUTE}?jwt_token={jwt_token}"
             )
+        else:
+            verification_link = f"https://{DOMAIN}:{PORT}{self._PARTICIPANTS_VERIFICATION_ROUTE}?jwt_token={jwt_token}"
```
Isak-Bego, line 342: Let's add a comment here that specifies that this format of the route is meant to be for local development

services/py-api/src/database/repository/participants_repository.py
```diff
@@ -36,26 +36,28 @@ async def fetch_by_id(self, obj_id: str) -> Result[Participant, ParticipantNotFo
...
+        #         doc_copy = dict(doc)
+        #
+        #         doc_copy["id"] = str(doc_copy.pop("_id"))
+        #
+        #         participants.append(Participant(**doc_copy))
+        #
+        #     LOG.debug(f"Fetched {len(participants)} participants.")
+        #     return Ok(participants)
+        #
+        # except Exception as e:
```
Isak-Bego, line 58: @KDimitrov333 Apparently the .find() method is not async in mongoDB for some reason. Can you please give it a check and create a PR with a fix in a new branch that is created from this branch.

services/py-api/src/database/repository/teams_repository.py
```diff
@@ -59,27 +59,28 @@ async def fetch_by_id(self, obj_id: str) -> Result[Team, TeamNotFoundError | Exc
...
+        #         doc_copy["id"] = str(doc_copy.pop("_id"))
+        #
+        #         teams.append(Team(**doc_copy))
+        #
+        #     LOG.debug(f"Fetched {len(teams)} teams.")
+        #     return Ok(teams)
+        #
+        # except Exception as e:
+        #     LOG.exception(f"Failed to fetch all teams due to err {e}")
+        #     return Err(e)
```
Isak-Bego, line 83: @KDimitrov333 Same thing applies here.

services/py-api/tests/unit_tests/test_database_layer/test_team_repository.py
```diff
@@ -206,57 +206,59 @@ async def test_fetch_by_id_general_error(db_manager_mock: Mock, repo: TeamsRepos
...
+#     result = await repo.fetch_all()
+#
+#     assert isinstance(result, Ok)
+#     assert len(result.ok_value) == 5
+#
+#     for i, team in enumerate(result.ok_value):
+#         assert team.name == mock_teams_data[i]["name"]
+#         assert team.is_verified == mock_teams_data[i]["is_verified"]
+#         assert team.created_at == mock_teams_data[i]["created_at"]
+#         assert team.updated_at == mock_teams_data[i]["updated_at"]
```
Isak-Bego, line 239: @KDimitrov333 and here

Isak-Bego (approved):
LGTM

PR #905, opened 2025-02-16
Journey section done

cl3vy:
Desktop:
image

Mobile:
image


DemonHashira (changes requested):
Good job! Still there will be changes xd. First is to actually put the text. You can get it from here: https://thehub-aubg.com/hackaubg  in the journey section.  We only need 4 and you can remove the rest

DemonHashira (changes requested):
Screenshot 2025-02-21 at 14 30 25

Another change is the mobile version. You can just make them flex-column and that's it. No need for the gsap here.

DemonHashira (dismissed):
Good work with GSAP. It's a hard one. Now that you have this checkpoint, can you make them stack on each other like it is in the design? Also make them start the line of the the title of the section.

Screenshot 2025-02-21 at 14 35 28
Screenshot 2025-02-21 at 14 35 40

Aleks1908 (approved):
ajde tedko

PR #918, opened 2025-02-20
General improvements to the API (in docs, structure, logic and tests)

IvanObreshkov:


services/py-api/src/app_factory.py
```diff
@@ -19,7 +20,7 @@ async def lifespan(app: FastAPI) -> AsyncIterator[None]:
     """
     yield
     # We ignore the error as we know that the client is initialized
-    _ = get_db_manager().close_all_connections()
+    _ = mongo_db_manager_provider(mongo_db_client_provider()).close_all_connections()
```
asynchroza, line 23: What are QoL of benefits of having these particular providers?

The seem to be really coupled with mongo. If there are plans to yield different provider instances (i.e. we use two databases which extend the same interface), this would make sense. In this particular case there is no clear benefit over the previous implementation. One might say that it even overcomplicates the code and makes it harder to read.

IvanObreshkov: There is a comment in the db_clients.py about that

asynchroza: Okay. A couple of notes:

- S3 is not a database. You will set yourself up for a failure if you want to come up with an extendable common interface which accommodates both object storages and databases. Generally, I don't believe the provider pattern is the best option for databases either, especially when your queries are not raw SQL .

- Premature optimization when the readability is not impacted is ok. If performance is improved but readability is impacted, that's ok in my book too (my current team at qb would kill you if the performance is not exceptionally impactful and the readability is impacted). But in this particular case, your implementation is just a more verbose version of what it was previously due to how coupled it is and not conforming to any interface contract (afair previous implementation initializes the client once in the module's scope and would always return the same reference to the client). IMHO converting to this non-correctly implemented provider pattern causes more confusion than anything else.

But either way, addressing this is totally up to you. I just believe codebase readability especially in the case of The Hub should be N.1 priority over anything else.

IvanObreshkov: I am up for a call when we both have more time. I am open to suggestions, as the idea of this PR (and the ones merged into it) is to do a general cleanup of what has been missed or not properly implemented due to time and other constraints.

PR #919, opened 2025-02-20
Fix jwt

IvanObreshkov:
* Pass the JwtUtility as a dependency so we can mock it
* Add Custom JWT schemas to improve type inference


Isak-Bego (approved):
LGTM! Good job!!!

Issue #922, opened 2025-02-22
Fix the team section from expanding


Isak-Bego:
## Brief description:
Since one of our members has a long name, it expands the card that it is placed, thus affecting the whole height of the carousel. 
## How to achieve it:
Find a solution to prevent this from happening: 
You can try ellipsis as @DemonHashira suggested

PR #924, opened 2025-02-22
922 Implements ellipses on hover and reworks the size dimensions for different devices

Alexandr-Ignatov:
#922 Implements ellipses on hover and reworks the size dimensions for different devices

## Added ellipses on hover effect for that one name:
Had to rework, the sizes of tiles and images, because otherwise it would either crop the image or would not display the full name on touch-screen devices.

Now it works on devices and it has an effect on hover


Aleks1908 (approved):
LGTM

PR #935, opened 2025-02-27
#934 implement FAQ section

### LINKED ISSUE #934: Implement the FAQ section
## Brief description:
Implement FAQ based on the Figma Design

Isak-Bego:
#934 implement FAQ section

## Implemented the FAQ section:
- Simple FAQ section that matches the Figma Design. 
- Keeps only one question open closing the previously opened ones
- Has an transition when opening the question
- its responsive

resolves #934

### DIFF
--- a/services/web/src/website/HackathonPage/FAQSection/HackathonFAQ.tsx
+++ b/services/web/src/website/HackathonPage/FAQSection/HackathonFAQ.tsx
@@ -1,4 +1,4 @@
-import { useState } from 'react';
+import { useState, useRef, useEffect } from 'react';
 
 interface FAQ {
     question: string;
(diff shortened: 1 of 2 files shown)


DemonHashira (changes requested):
Top work, brat! Only two changes. The first is to make the opening and closing smooth. The transition property should be able to do it.

DemonHashira (changes requested):
Screenshot 2025-02-28 at 9 20 19

Also, on mobile, I think we should cut the padding a bit. In my opinion, there's too much space between the footer and the FAQ xd.

Isak-Bego:
> Screenshot 2025-02-28 at 9 20 19
> 
> Also, on mobile, I think we should cut the padding a bit. In my opinion, there's too much space between the footer and the FAQ xd.

I believe this could be due to chrome dev tools in responsive mode. They add more padding then there actually is. Here is what firefox which has a more accurate responsive design tool gives me: 
image

DemonHashira:
> > Screenshot 2025-02-28 at 9 20 19
> > Also, on mobile, I think we should cut the padding a bit. In my opinion, there's too much space between the footer and the FAQ xd.
> 
> I believe this could be due to chrome dev tools in responsive mode. They add more padding then there actually is. Here is what firefox which has a more accurate responsive design tool gives me: image

Hmmm. That might be the case. We will see how it is when we upload it on dev and can see on real devices.

PR #944, opened 2025-03-01
Split nginx docker files to prod and dev and add more docs

IvanObreshkov:
Split the web images into prod and dev ones, in order to use different nginx config and variables, baked into the image


.github/workflows/dev_deploy.yml
```diff
@@ -58,14 +58,15 @@ jobs:
...
+          file: ./services/web/dev.Dockerfile
           push: true
           tags: |
             ghcr.io/aubgthehub/monolith-web:latest
             ghcr.io/aubgthehub/monolith-web:${{ github.sha }}
           # We pass the build arguments for the frontend
           # https://docs.docker.com/reference/cli/docker/buildx/build/#build-arg
           build-args: |
-            VITE_ENV=${{ vars.VITE_ENV }}
+            DOMAIN=${{vars.DOMAIN}}
```
Isak-Bego, line 69: We should not change this without consulting with @Aleks1908 . Since we are building the frontend to static files this argument should be passed with the correct name `VITE_ENV` so that its correctly recognized by vite and properly substituted with its value in the js bundle

services/web/dev.Dockerfile
```diff
@@ -0,0 +1,47 @@
...
+COPY . .
+
+# Install dependencies (npm ci makes sure the exact versions in the lockfile gets installed)
+RUN npm ci
+
+ARG DOMAIN
+
+# Set environment variable based on build argument. These are passed in the CD pipeline.
+# The environment variables set using ENV will persist when a container is run from the resulting image.
+ENV VITE_ENV="PROD"
```
Isak-Bego, line 18: Okay so we are hardcoding the values of VITE_ENV instead of passing it through the action since we now have 2 separate docker files. This value should be "DEV" then.

services/web/prod.Dockerfile
```diff
@@ -0,0 +1,47 @@
...
+
+# Install dependencies (npm ci makes sure the exact versions in the lockfile gets installed)
+RUN npm ci
+
+ARG DOMAIN
+
+# Set environment variable based on build argument. These are passed in the CD pipeline.
+# The environment variables set using ENV will persist when a container is run from the resulting image.
+ENV VITE_ENV="PROD"
+ENV DOMAIN=$DOMAIN
```
Isak-Bego, line 19: What are we using this ENV for?

IvanObreshkov: The health checks

Isak-Bego (approved):

PR #945, opened 2025-03-02
#921 Update email templates

### LINKED ISSUE #921: Restructure the email templates
## Brief description:
Since we had troubles with the email templates last year because of the way they are handled by the email clients, we have decided to simplify their structure this year. This will not only lower the chances that they end up in spam folders, but also ensure consistency. 
## How to achieve it:
You can find the templates in py-api>src>services>mail_service>email_templates
From there you can start editing their structure, so that it matches the picture that is attached. From now we can keep the same words and keep the placeholders.

Image

asiahaslam:
resolves #921


asynchroza:
There's a lot of funky stuff going on in this PR. What are all the new additions?

Isak-Bego:
> There's a lot of funky stuff going on in this PR. What are all the new additions?

The build files were included in the commit thats why

Isak-Bego (approved):
Looks good to me!

PR #949, opened 2025-03-07
remove embla carousel from past events section

Isak-Bego:
## Removed the Carousel:
After multiple days of debugging the carousel, I decided to switch to a simple div section until we migrate to a lighter carousel implementation with better compatibility. 
The was an animation stutter on IOS specifically for this section that I could not figure out, which apparently was reported by other developers as well. However, the maintainer of this library has not given a clear explaination to why this is happening yet. 
I also converted the pictures that we are using for the cards to .webp for better compression and loading speed


DemonHashira (approved):
Top G

PR #962, opened 2025-03-10
#923 fix the gradient blobs across all browsers

### LINKED ISSUE #923: fix the gradient blobs across all browsers
## Brief description:
We are having trouble with they way the gradient works on firefox on Mac. To fix this issue we will no longer apply the gradient on the browser but rather insert a png of the gradient in the site, so that it works the same for all browsers.
## How to achieve it:
Download the pngs from figma. place them in the same spots where the svgs are now. Try to keep the same design as it is now.

Bubomira:
# gradient blobs replaced with a png version of the svgs
#resolves #923


Aleks1908 (approved):
LGTM

PR #968, opened 2025-03-11
967 Make Registration Form Wider and Add Missing Options

### LINKED ISSUE #967: Make Registration Form Wider and Add missing options
The width of the form should be more and the dropdown "Source of referral" should have the option "Social Media". For this to work the option should be added in both the frontend and backend

MagdalenaZh:
#967 Make Registration Form Wider and Add missing options

Made registration form wider and added missing "Social media" option

resolves #967


Aleks1908 (approved):
ЛГТМ

PR #985, opened 2025-03-12
Fixing Front end nits

Aleks1908:
What has been fixed: 
- About button on HackAubg page was not working
- Navbar did not have background on the RegPage
- Whitelist dashes for name input field
- Add date check so that we have appropriate messages for when the registration form is not open
- Add new video


Isak-Bego (approved):
LGTM

Issue #978, opened 2025-03-12
Remove XS T-shirt option


Aleks1908:
Remove XS T-shirt option from the registration form and remove it from the supported literal in the backend

PR #986, opened 2025-03-12
Remove XS T-shirt option

MagdalenaZh:
#978  Remove XS T-shirt option

Removed XS T-shirt option from the registration form and the backend.


Aleks1908 (approved):
LGTM

Issue #946, opened 2025-03-04
Docker containers restarting on failing healtchecks and get stuck in a restarting loop.


IvanObreshkov:
## Brief description:
During testing on DEV with @Isak-Bego and @asynchroza we discovered an issue where the container where the containers were not passing the health check and were restarting after that forever. Not sure if this is the expected behavior of the [`restart_policy`] (https://github.com/docker/compose/blob/v1/docs/Compose%20file%20reference%20(legacy)/version-3.md#restart_policy) in the `docker-swarm-stack.yaml`

## Additional issues:

We should reconsider the placements of the health checks (whether they should be inside the docker file and baked into the image, or passed when running the container via the compose file)

PyAPI has an incorrect HEALTHCHECK, as it depends on whether nginx is up or not (the check is a curl towards https://dev.thehub-aubg.com/api/v3/ping , and if nginx is not up, the health check will fail, even if PY-API is healthy itself)

## How to achieve it

Check the Dockerfiles in `web` and `py-api`, as well as the `docker-swarm-stack.yaml` file

PR #995, opened 2025-03-16
Optimize docker images

IvanObreshkov:
Fix: #946 

* Added folder structure for docs
* Optimized both WEB and PY-API images, by:
* * utilizing [cache-mounts](https://docs.docker.com/build/ci/github-actions/cache/#cache-mounts)
* * removing health checks from the images, and placing them in the compose specification
* * Run the PY-API container with a non-root user as per: https://docs.docker.com/build/building/best-practices/#user
* * Added more docs regarding the self-heal mechanism of docker swarm

### Note:
* We need to change the deployment pipeline to utilize the cache-mounts, by using [buildkit-cache-dance](https://github.com/reproducible-containers/buildkit-cache-dance)
* We need to change nginx to use the non-root user.
* This PR should be merged when we finish refactoring the code.


services/py-api/Dockerfile
```diff
@@ -1,44 +1,65 @@
...
-    POETRY_VIRTUALENVS_IN_PROJECT=1 \
-    POETRY_VIRTUALENVS_CREATE=1 \
+ENV POETRY_VIRTUALENVS_IN_PROJECT=1 \
     POETRY_CACHE_DIR=/tmp/poetry_cache
 
 WORKDIR /app
 
 # Poetry will complain if a README.md is not found
-COPY pyproject.toml poetry.lock README.md ./
+RUN touch README.md
```
asynchroza, line 13: Lol

services/web/dev.Dockerfile
```diff
@@ -1,30 +1,46 @@
...
 ENV VITE_ENV="DEV"
 
 # Build the app
 RUN npm run build
 
 # Bundle static assets with nginx
-FROM nginx:latest AS dev
+# TODO: Change image with https://hub.docker.com/r/nginxinc/nginx-unprivileged when the auto renew of ssl certs is
+# implemented, as oer: https://docs.docker.com/build/building/best-practices/#user
+FROM nginx:1.27-alpine AS dev
```
asynchroza, line 34: Why do you need a dev dockerfile?

IvanObreshkov: cause we are copying a different nginx configs depending on the env, and I didn't want to do if-else copy. (COPY does not allow it, and I had to use RUN)

asynchroza: I'm not sure I understand what you mean by exposing too many details. Changes in directory structure would always mean editing the path pointing to the configuration - no matter whether it was defined as a build argument or hardcoded. 

I doubt that your autocomplete extension is firing off errors if path is incorrect. Hence, having the path hardcoded there is not of any net benefit.

You could possibly introduce a linter to catch these errors but that's an overkill. Of course that changes in structure would dictate changes in configurations. But do you really expect too much of a change in structure here?

IvanObreshkov: Tbh, not really. So, you recommend either hardcoding it in the CD pipeline or extracting it as a GitHub ENV var, right? I prefer the latter if that's the case.

asynchroza: I recommend conditionally inferring it in the deployment pipeline

If (branch === production) prefix = "prod." else "";
arg PATH_TO_NGINX_FILE = path + prefix + file

IvanObreshkov: @asynchroza Fixed in [58d6367](https://github.com/AUBGTheHUB/monolith/pull/995/commits/58d63675609ce780d6fe3df3f761351744fb2734)

PR #1001, opened 2025-03-24
Fastapi without naitve di

IvanObreshkov:


services/py-api/src/app_factory.py
```diff
@@ -1,32 +1,100 @@
...
     """
+    # Open a connection to Mongo
+    db_client = mongo_db_client_provider()
+
     yield
-    # We ignore the error as we know that the client is initialized
-    _ = mongo_db_manager_provider(mongo_db_client_provider()).close_all_connections()
+
+    # Close the opened connections towards MongoDB
+    db_client.close()
```
asynchroza, line 56: I know this is straightforward for you but I suggest you throw a quick comment explaining when is the yield invoked and what it does

services/py-api/src/server/routes/feature_swithces_routes.py
```diff
@@ -0,0 +1,30 @@
...
+def register_feature_switches_routes(router: APIRouter, http_handler: FeatureSwitchHandler) -> None:
+    router.add_api_route(
+        path="/feature-switches/{feature}",
+        methods=["GET"],
+        endpoint=http_handler.get_feature_switch,
+        responses={200: {"model": FeatureSwitchResponse}, 404: {"model": ErrResponse}},
+    )
+
+    router.add_api_route(
+        path="/feature-switches",
```
asynchroza, line 17: Not sure how this works here but something I've seen elsewhere is creating a router, chain all the paths to the router created within the function, return the router and bind it under the feature path I.e. /feature-switches. This way you could avoid potential issues with incorrect naming. If that's not feasible in your case, extract the parent path as a constant.

asynchroza (commented):
Lgtm. No concerns besides the one we've already talked about.

services/py-api/src/server/routes/feature_swithces_routes.py
IvanObreshkov: SO you suggest having something like this:
```python
def register_feature_switches_routes(router: APIRouter, http_handler: FeatureSwitchHandler) -> None:
    fs_router = APIRouter(prefix="/feature-switches")

    fs_router.add_api_route(
        path="/{feature}",
        methods=["GET"],
        endpoint=http_handler.get_feature_switch,
        responses={200: {"model": FeatureSwitchResponse}, 404: {"model": ErrResponse}},
    )

    fs_router.add_api_route(
        path="",
        methods=["PATCH"],
        endpoint=http_handler.handle_feature_switch_update,
        responses={200: {"model": FeatureSwitchResponse}, 404: {"model": ErrResponse}, 400: {"model": ErrResponse}},
        dependencies=[Depends(is_auth)],
    )

    fs_router.add_api_route(
        path="",
        methods=["GET"],
        endpoint=http_handler.get_all_feature_switches,
        responses={200: {"model": AllFeatureSwitchesResponse}},
    )

    # Bind the fs_router (request multiplexer) to the main router
    router.include_router(fs_router)
```

Issue #816, opened 2024-12-05
Add TraceID to logging


IvanObreshkov:
## Brief description:
You said you could set this up for 5 min, that's why I am assigning it to you :D. 

As discussed, it's a good idea to add traceID to our logs so that we can trace the operations happening when it comes to handling a request. For this to work, maybe we would need the uvicorn log entries to be in a JSON format as well when it comes to deployed environments.

## How to achieve it:
I am not sure but I guess you should do your magic in this file:
https://github.com/AUBGTheHUB/monolith/blob/main/services/py-api/src/server/logger/logger_factory.py 

Maybe you could get some inspiration from here, or maybe not:
https://www.w3.org/TR/trace-context/

Ti si znaesh brat.

IvanObreshkov:
https://www.structlog.org/en/stable/contextvars.html
This might be helpful

PR #1110, opened 2025-11-23
#816 Add per-request trace ID to app and uvicorn logs

SiyanaKostova:
## Summary

Add a per-request UUID trace ID and wire it into both application logs (structlog) and uvicorn logs (in JSON format for deployed environments), so we can trace everything that happened for a given request.

## Changes

- Add `register_request_id_middleware`:
  - Generates or reuses `X-Request-ID` (UUID hex).
  - Stores it on `request.state.request_id`.
  - Binds it into structlog contextvars as `request_id`.
  - Ensures the response always includes `X-Request-ID`.

- Update `logger_factory.get_uvicorn_logger`:
  - Add `RequestIdFilter` that pulls `request_id` from structlog contextvars into the log record.
  - For `ENV=DEV` and `ENV=PROD`, log uvicorn output to `shared/server.log` as JSON-like lines including `"request_id"`.

## Testing

- Locally (ENV=LOCAL), hit:

  ```bash
  curl -vk https://localhost:8080/api/v3/ping -i


services/py-api/src/server/middleware/request_id_middleware.py
```diff
@@ -0,0 +1,55 @@
...
+    @app.middleware("http")
+    async def request_id_middleware(
+        request: Request,
+        call_next: Callable[[Request], Awaitable[Response]],
+    ) -> Response:
+        # Make sure we start from a clean context for this request
+        clear_contextvars()
+
+        # Reuse incoming header if present, otherwise generate a fresh UUID
+        request_id = request.headers.get(REQUEST_ID_HEADER) or uuid.uuid4().hex
```
IvanObreshkov: 1. Currently, we don't have a use case where an upstream service sends us an `X-Request-ID` request header.
2. The convention of using `X-` has been deprecated: https://datatracker.ietf.org/doc/html/rfc6648
3. With the current setup of Caddy, we pass through all headers sent from client services. This means that a malicious actor could inject a wrongful request ID, and especially if we use this request for some logic in the downstream, we could get a problem.

If Caddy generated such headers for us, this would be amazing, but it does not :D See: https://caddy.community/t/generate-a-unique-request-id-when-using-a-reverse-proxy/19519/2

It only uses the `traceID`, which is for OpenTelemetry, but we are not instrumenting the Hub website any time soon, so yeah. See: `https://caddy.community/t/generate-a-unique-request-id-when-using-a-reverse-proxy/19519/2

If you don't have a valid reason to use this header, please remove it.

services/py-api/src/server/middleware/request_id_middleware.py
```diff
@@ -0,0 +1,55 @@
...
+
+        try:
+            response = await call_next(request)
+        finally:
+            # Avoid leaking the request_id into unrelated contexts
+            unbind_contextvars(REQUEST_ID_CTX_KEY)
+
+        # Make sure clients can see / report this ID back if needed
+        if REQUEST_ID_HEADER not in response.headers:
+            response.headers[REQUEST_ID_HEADER] = request_id
```
IvanObreshkov: This is cool, but please revisit the name of the header. Example: `TheHubAUBG-Request-ID`

services/py-api/src/logger/logger_factory.py
```diff
@@ -110,6 +144,7 @@ def get_uvicorn_logger(env: str) -> Dict[str, Any]:
                 "formatter": "logformatter",
                 "maxBytes": 10 * 1024 * 1024,  # 10 MB limit
                 "backupCount": 2,  # 2 backup files
+                "filters": ["request_id_filter"],
```
IvanObreshkov: Apparently, this thing does not work :(

TBH, I have no idea what these filters are supposed to do, so I cannot help without digging deeper.

Here are the structured logs in JSON (When we are running in deployed env)
```json
{"event": "Pinging MongoDB...", "level": "debug", "request_id": "93290fb6b7b4462cb1f01b2aa8ea57de", "timestamp": "2025-11-23 17:09:04", "func_name": "async_ping_db", "lineno": 37, "filename": "db_manager.py"}
{"event": "Pong", "level": "debug", "request_id": "93290fb6b7b4462cb1f01b2aa8ea57de", "timestamp": "2025-11-23 17:09:04", "func_name": "async_ping_db", "lineno": 39, "filename": "db_manager.py"}
{"timestamp":"2025-11-23T19:09:04","level":"INFO","logger":"uvicorn.access","request_id":"","message":"127.0.0.1:58383 - "GET /api/v3/ping HTTP/1.1" 200"}
```

And here are the logs in local env:

Image

services/py-api/src/logger/logger_factory.py
```diff
@@ -93,13 +98,42 @@ def doRollover(self) -> None:
...
+    based on structlog's contextvars.
+    """
+
+    def filter(self, record: LogRecord) -> bool:
+        try:
+            ctx = get_contextvars()
+        except Exception:
+            ctx = {}
+
+        request_id = ctx.get("request_id")
```
IvanObreshkov: You used a const earlier for this key; why not use it again? I know that dictionaries are not the best thing in the world, as they are not type-safe, but since you don't really know what you could have in the context variables for a logger and you want to put many keys in this context, this is why they have used a dictionary.

It would be even better to create a TypedDict above `RequestIdFilter` and list the keys we have in the structlog contextvars. In this case, it is going to be even better:

```python
class StruclogContextVars(TypedDict):
    request_id: str
...

ctx: StruclogContextVars = get_contextvars()
# You will get autocomplete on the keys in the dict when you try to reference them
request_id = ctx['request_id']
```

IvanObreshkov (changes requested):

services/py-api/src/app_factory.py
```diff
@@ -99,6 +100,7 @@ async def lifespan(app: FastAPI) -> AsyncIterator[None]:
 def create_app() -> FastAPI:
     app = FastAPI(lifespan=lifespan, root_path="/api/v3")
 
+    register_request_id_middleware(app)
```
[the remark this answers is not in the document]
IvanObreshkov: @SiyanaKostova Fixed in [6bb1870](https://github.com/AUBGTheHUB/monolith/pull/1110/commits/6bb18702ea40588a1604e0cb8a348c04dccbf2d7) and [5a7008f](https://github.com/AUBGTheHUB/monolith/pull/1110/commits/5a7008f391db34f61c14150b10280446a660e9c2)

services/py-api/src/logger/logger_factory.py
IvanObreshkov: @SiyanaKostova Fixed in [3a024bf](https://github.com/AUBGTheHUB/monolith/pull/1110/commits/3a024bf986b3b769e4327f6b1bcb635a3ebd3bea)

services/py-api/src/logger/logger_factory.py
IvanObreshkov: Fixed in [279a82d](https://github.com/AUBGTheHUB/monolith/pull/1110/commits/279a82d5f61fd4f428d8a4f4f35d32fea64d7f87)

PR #1131, opened 2026-01-04
Create scaffolding for auth

georginikolov7:
### Description

Created endpoints:
  - auth/login
  - auth/register
  - auth/refresh
  
 Created auth handlers and service

Created admin model

### DIFF
--- a/services/py-api/src/server/routes/auth/auth_routes.py
+++ b/services/py-api/src/server/routes/auth/auth_routes.py
@@ -8,17 +8,17 @@ def register_auth_routes(http_handler: AuthHandlers) -> APIRouter:
     and returns the router"""
 
     # TODO ADD responses to each path according to prior project structure
-    authentication_router = APIRouter(prefix="/auth", tags=["auth"])
-    authentication_router.add_api_route(
+    auth_router = APIRouter(prefix="/auth", tags=["auth"])
+    auth_router.add_api_route(
         path="/login",
         methods=["POST"],
         endpoint=http_handler.login,
     )
-    authentication_router.add_api_route(
+    auth_router.add_api_route(
         path="/register",
         methods=["POST"],
         endpoint=http_handler.register,
     )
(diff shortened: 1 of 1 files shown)


IvanObreshkov (commented):
Please remove the .idea folder as well. Aside from that, it looks fine for now

services/py-api/src/database/model/admin/hub_member_model.py
```diff
@@ -6,6 +6,7 @@
 from src.database.model.base_model import BaseDbModel, UpdateParams
 
 DEPARTMENTS_LIST = Literal["Development", "Marketing", "Logistics", "PR", "Design"]
+MEMBER_TYPE = Literal["member", "admin"]
```
IvanObreshkov: Add a docstring about this, regarding the distinction between admin and member. Which HubMembers are admins?
```python
MEMBER_TYPE = Literal["member", "admin"]
"""Some description"""
```

services/py-api/src/server/routes/route_dependencies.py
```diff
@@ -25,6 +25,7 @@
 # ===============================
 
 
+# TODO Update with new auth
 def is_auth(authorization: Annotated[str, Header()]) -> None:
```
IvanObreshkov: Please rename this to `is_authorized`, to be more explicit.

services/py-api/src/database/model/admin/hub_member_model.py
georginikolov7: Exactly right. Will describe in docstring

services/py-api/src/server/handlers/auth/auth_handlers.py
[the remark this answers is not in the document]
georginikolov7: Agreed. Will fix

georginikolov7:
Btw the pre-commit autoformatting seems to have unprettified some route configs if you notice so thats not on me

IvanObreshkov (approved):

PR #1141, opened 2026-01-08
Create setup scripts

georginikolov7:
### NOT TESTED ON MacOS. Somebody on Mac please take a look.

Create install and uninstall scripts for local installation of dependancies and update readme file

Running install_depts.sh installs gum, pyenv, python 3.12.9, npm, poetry and all libraries
Running uninstall_depts.sh will uninstall everything that install installed. 

#### WARNING:
Running unistall will delete all dependencies so carefull as you might have them previously istalled for another project.

### DIFF
--- a/setup/install_depts.sh
+++ b/setup/install_depts.sh
@@ -3,26 +3,46 @@
 echo "🚀 Starting environment setup..."
 
 OS_TYPE="$(uname)"
+# Detect Linux distribution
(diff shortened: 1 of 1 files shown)


IvanObreshkov:
@Aleks1908 or @DemonHashira, could any of you guys please test this, as I have those tools already installed locally on my machine?

IvanObreshkov (approved):

PR #1150, opened 2026-01-14
Added crud for sponsors in the admin panel

StoyanChorbov:
All functionality has been implemented and manually tested. Still need to do integration and unit tests, so maybe don't close the branch yet.

### DIFF
--- a/services/py-api/tests/unit_tests/test_database_layer/test_sponsors_repository.py
+++ b/services/py-api/tests/unit_tests/test_database_layer/test_sponsors_repository.py
@@ -69,13 +69,13 @@ async def test_create_sponsor_general_exception(
 @pytest.mark.asyncio
 async def test_delete_sponsor_success(
     mongo_db_manager_mock: MongoDbManagerMock,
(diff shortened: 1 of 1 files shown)


services/py-api/src/server/handlers/admin/sponsors_handlers.py
```diff
@@ -2,23 +2,55 @@
...
+        if is_err(result):
+            return self.handle_error(result.err_value)
+
+        return Response(SponsorResponse(sponsor=result.ok_value), status_code=201)
 
     async def get_all_sponsors(self) -> Response:
-        raise NotImplementedError()
+        result = await self._service.get_all()
+
+        print(result)
```
georginikolov7: Again why leave this in the code?

StoyanChorbov: Like I said, I just forgot it was there, or this is a commit that was made before I realized I have to remove it. P.S. I didn't realize the logger's output was the terminal, as I am used to it being a random section in the IDE or a weird external console.

Issue #1133, opened 2026-01-04
Implement Image Fetching Logic


georginikolov7:
## Brief description:
As discussed on the meeting we need to implement a service that downloads/uploads images to aws

### Create entity (past_event, mentor etc):
1. FE → BE
- Request a pre-signed upload URL

2. BE → AWS S3
-Generates pre-signed URL
-Returns it to React

3. FE → S3
-Uploads image directly to S3 

4. FE → BE
- Save model to mongoDB. 
!IMPORTANT: if anything fails ->delete the saved image from aws s3

### Update entity:
1.  FE -> BE
 - if img_url is different upload the new image
 - PUT request to be to update entitiy
 
2.BE
- try to update entity
- on success -> delete old image
- on fail -> delete new image and return corresponding error

PR #1184, opened 2026-01-27
#1133 Image upload logic

StoyanChorbov:
Added logic to upload images to the AWS S3 bucket along with tests.


services/py-api/src/service/utility/image_storing/image_storing_service.py
```diff
@@ -0,0 +1,79 @@
...
+        except Exception as e:
+            LOG.exception("There was an error when uploading the image", error=e)
+            raise ImageUploadError() from e
+
+    def delete_image(self, file_name: str) -> None:
+        try:
+            self._aws_service.delete_file(file_name)
+
+        except Exception as e:
+            LOG.exception("There was an error when deleting the image", error=e)
```
georginikolov7: Isn't it by url

PR #1186, opened 2026-01-28
Implement generic apiClient and wire up sponsors

georginikolov7:
## Feats:
 1. Implemented generic apiClient.ts file for fetching data from BE. File includes GET, PUT, PATCH,DELETE endpoint requests and all currently needed headers.
 2. Use the implemented apiClient in combination with the **TanStackQuery library** to fetch Sponsors from BE. Admin Panel Sponsors page is now functional. 
 3. Temporarely removed careers_url as it still does not exist on BE. Will get re-added.

### DIFF
--- a/services/web/src/internalLibrary/AdminCard/adminCard.tsx
+++ b/services/web/src/internalLibrary/AdminCard/adminCard.tsx
@@ -29,18 +29,10 @@ export function AdminCard({
     ...rest
 }: AdminCardProps) {
     return (
-        <Card 
-            className={cn('overflow-hidden flex flex-col h-full border-0', className)}
-            {...rest}
-        >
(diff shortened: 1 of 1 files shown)


services/web/src/website/AdminPanelPage/CRUDPages/SponsorsPage/validation/sponsor.tsx
```diff
@@ -8,7 +8,7 @@ const RULES = {
 export const sponsorSchema = z.object({
     name: z
         .string()
-        .min(1, { message: 'Sponsor name is required' })
+        .min(1, { message: 'Validation name is required' })
```
Aleks1908: Why did you change this?

PR #1187, opened 2026-01-28
Implement VerticalBar component

georginikolov7:
## Description 
Implemented vertical bar component which takes two boolean variables: isRight and isBlack, to determine the position and color.
Component is fully reusable in all sections

Note: Also fixed some issues with FAQ Section:
- add min-h-dvh so it takes up one screen
- add left and right margin to FAQ section


DemonHashira (dismissed):
Aide bratio

PR #1190, opened 2026-01-29
#1082 authentication for admin panel 

### LINKED ISSUE #1082: Implement Admin Authentication
## Description
Implement BE authorization to protect the admin panel endpoints
- There will be three admin roles -  board, dev, super_admin. Only super_admin can assign roles.
- Registration endpoint creates hub_admin_member models
- Follow outh 2.0 standard
- Use the provided auth sceleton
- Authentication should use username and password system

### Note:
- There exist a hub_admin_member model that inherits hub_member. Mongo allows for admins to still be imported into the hub_member document, so use that feature for simplicity. 
-  Registration endpoint will be on a feature switch so that it is only live until all board members register. Do not worry about this functionality for now.

### Requirements:
- Passwords are stored in mongoDb HASHED!!!
- Bearer token is encrypted
- All endpoints in sceleton are implemented
- Full unit and integration testing

Bubomira:
# Implement basic authentication
closes #1082 
- auth routes implemented
- created request/response schemas
- created auth handler that supports register, login and refresh token
- created auth service that  supports register, login and refresh token
- created password_hash_service with bcrypt to store passwords hashed in db
- created auth_token_service to issue auth tokens and refresh tokens, and to check refresh tokens
- implemented hub_members repository working for both admins and hub members
- implemented refresh tokens repository for the refresh token entity
- implemented appropriate JWT schemas for the auth and refresh tokens
- both unit and integration tests
### REMARK 
One of the integration tests does not run because we haven't created an unique index in mongodb from compass for the usernames of hub admins

### DIFF
--- a/services/py-api/src/service/auth/auth_service.py
+++ b/services/py-api/src/service/auth/auth_service.py
@@ -36,6 +36,32 @@ class AuthService:
         self._refresh_token_repo = refresh_token_repo
         self._tx_manager = tx_manager
 
+    async def _invalidate_old_and_create_new_refresh_token_callback(
+        self, refresh_token_id: str, refresh_expiration: datetime, session: Optional[AsyncIOMotorClientSession] = None
+    ) -> Result[RefreshToken, Exception]:
+
+        # Invalidate the old refresh token
+        updated_token_result = await self._refresh_token_repo.update(
+            refresh_token_id, obj_fields=UpdateRefreshTokenParams(is_valid=False), session=session
+        )
+
+        if is_err(updated_token_result):
+            return updated_token_result
+
+        hub_admin_id = updated_token_result.ok_value.hub_member_id
+        family_id = updated_token_result.ok_value.family_id
+
+        # Create new refresh token from this family id
+        new_refresh_token = RefreshToken(
+            hub_member_id=hub_admin_id, family_id=family_id, is_valid=True, expires_at=refresh_expiration
+        )
+        new_token_result = await self._refresh_token_repo.create(new_refresh_token, session=session)
+
+        if is_err(new_token_result):
+            return new_token_result
+
+        return Ok(new_token_result.ok_value)
+
     async def login_admin(
         self, credentials: LoginHubAdminData
     ) -> Result[tuple[str, str], HubMemberNotFoundError | PasswordsMismatchError | Exception]:
@@ -160,29 +186,3 @@ class AuthService:
         )
 
         return Ok((jwt_auth_token, jwt_refresh_token))
-
-    async def _invalidate_old_and_create_new_refresh_token_callback(
-        self, refresh_token_id: str, refresh_expiration: datetime, session: Optional[AsyncIOMotorClientSession] = None
-    ) -> Result[RefreshToken, Exception]:
-
-        # Invalidate the old refresh token
-        updated_token_result = await self._refresh_token_repo.update(
-            refresh_token_id, obj_fields=UpdateRefreshTokenParams(is_valid=False), session=session
-        )
-
-        if is_err(updated_token_result):
-            return updated_token_result
-
-        hub_admin_id = updated_token_result.ok_value.hub_member_id
-        family_id = updated_token_result.ok_value.family_id
-
-        # Create new refresh token from this family id
-        new_refresh_token = RefreshToken(
-            hub_member_id=hub_admin_id, family_id=family_id, is_valid=True, expires_at=refresh_expiration
-        )
-        new_token_result = await self._refresh_token_repo.create(new_refresh_token, session=session)
-
-        if is_err(new_token_result):
-            return new_token_result
-
-        return Ok(new_token_result.ok_value)


services/py-api/src/database/model/admin/hub_member_model.py
```diff
@@ -28,14 +28,24 @@ class HubMember(BaseDbModel):
     avatar_url: str
     social_links: SocialLinks = field(default_factory=lambda: cast(SocialLinks, cast(object, {})))
 
+    def _serialize_social_links(self) -> dict[str, Any]:
+        social_links_serialized = {}
+        for link_name, link_url in self.social_links.items():
+            if link_url is not None:
+                social_links_serialized[link_name] = str(link_url)
+
+        return social_links_serialized
```
IvanObreshkov: Why do we need this? Wouldn't the type `HttpUrl` be enough to prevent/specify that we do not expect to have a nullable value for a link_url

```python
class SocialLinks(TypedDict):
    linkedin: NotRequired[HttpUrl]
    github: NotRequired[HttpUrl]
    website: NotRequired[HttpUrl]
```

services/py-api/src/exception.py
```diff
@@ -132,6 +134,26 @@ class FeatureSwitchNotFoundError(CustomError):
     status_code = status.HTTP_404_NOT_FOUND
 
 
+class HubMemberNotFoundError(CustomError):
+    message = "The HUB member was not found."
+    status_code = status.HTTP_404_NOT_FOUND
+
+
+class DuplicateHUBMemberNameError(CustomError):
```
IvanObreshkov:
```python
class DuplicateHubMemberNameError(CustomError):
```

services/py-api/src/service/auth/password_hash_service.py
```diff
@@ -0,0 +1,21 @@
+from typing import cast
+from bcrypt import checkpw, gensalt, hashpw
+
+SALT_ROUNDS = 20
```
IvanObreshkov: Why 20 rounds? Bcrypt is a fairly slow algorithm, so increasing the number of rounds will increase the response times quite a lot. I mean, registering users is not an operation that happens often, so we should be fine, but please add a docstring explaining why we have chosen 20. The default is 12, so this 66% increase should be justified.

services/py-api/src/service/auth/password_hash_service.py
```diff
@@ -0,0 +1,21 @@
...
+
+class PasswordHashService:
+    def __init__(self) -> None:
+        pass
+
+    def hash_password(self, password_string: str) -> bytes:
+        bytes_pass = password_string.encode("utf-8")
+        salt = gensalt(SALT_ROUNDS)
+
+        return cast(bytes, hashpw(password=bytes_pass, salt=salt))
```
IvanObreshkov: Why cast? 

Image

services/py-api/src/service/auth/password_hash_service.py
```diff
@@ -0,0 +1,21 @@
...
+from bcrypt import checkpw, gensalt, hashpw
+
+SALT_ROUNDS = 20
+
+
+class PasswordHashService:
+    def __init__(self) -> None:
+        pass
+
+    def hash_password(self, password_string: str) -> bytes:
```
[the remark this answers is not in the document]
IvanObreshkov: Now, to the biggest issue with this code. Hashing passwords is a CPU-bound operation. This means that when you execute this function, you will block the whole thread on which the [AsyncIO event loop](https://docs.python.org/3.12/library/asyncio-eventloop.html#asyncio-event-loop) is running on. [Only one event loop can run on a given thread.](https://docs.python.org/3.12/library/asyncio-policy.html#asyncio-policy-objects)

The good thing for us is that the bcrypt library [releases the GIL](https://github.com/pyca/bcrypt/blob/f0451e42e3ab6f6e1b9ac8b09bf04104bf8bdef8/src/_bcrypt/src/lib.rs#L193), meaning we can confidently use the `asyncio.to_thread`   method for this CPU-bound operation, as written in the [docs](https://docs.python.org/3.12/library/asyncio-task.html#asyncio.to_thread). You can read more about the GIL [here](https://realpython.com/python-gil/).

Side note:

Here we don't care about the hashing operation timing out, but if, for some reason, you need to have a cancellation mechanism due to some time constraints or not wanting to wait indefinitely for the result from the operation, which you pleased in the `to_thread` func, you can provide a timeout via [wait_for](https://docs.python.org/3/library/asyncio-task.html#asyncio.wait_for). Keep in mind that `wait_for` will cancel the future returned (not the thread), but unfortunately, the thread will continue running indefinitely, which could create strange behavior. You can check more https://github.com/python/cpython/issues/87185. Instead, you should be using [asyncio.loop.run_in_executor](https://docs.python.org/3.12/library/asyncio-eventloop.html#asyncio.loop.run_in_executor) to schedule the background task and get a future you can use wait_for to wait on the event-loop.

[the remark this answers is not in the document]
IvanObreshkov: Also, I don't know what has happened, but currently the Py-Api tests are running on your branch for almost 7 min:

Image

And previously they have been running for two:
https://github.com/AUBGTheHUB/monolith/actions/runs/21531309578/job/62047629469

I am surprised that your 4-5 new tests create an additional 5 min. Please investigate where the latency comes from. It could be related to the blocking of the event loop.

services/py-api/src/service/auth/auth_token_service.py
```diff
@@ -0,0 +1,26 @@
...
+from src.database.model.admin.hub_admin_model import HubAdmin
+from src.service.jwt_utils.codec import JwtUtility
+from src.service.jwt_utils.schemas import JwtAdminToken, JwtRefreshToken
+
+
+class AuthTokenService:
+    def __init__(self, jwt_utility: JwtUtility):
+        self._jwt_utility = jwt_utility
+
+    def generate_auth_token(self, hub_admin: HubAdmin) -> str:
```
IvanObreshkov: The term used in the [OAuth2.0 Spec](https://datatracker.ietf.org/doc/html/rfc6749#section-4.2.2) is `access_token`, so please change this to be:

```python
def generate_access_token_for()
```

services/py-api/src/service/auth/auth_token_service.py
```diff
@@ -0,0 +1,26 @@
...
+        expiration = int((datetime.now(timezone.utc) + timedelta(minutes=7)).timestamp())
+        payload = JwtAdminToken(
+            sub=str(hub_admin.id), exp=expiration, site_role=hub_admin.site_role, member_type=hub_admin.member_type
+        )
+        return self._jwt_utility.encode_data(data=payload)
+
+    def generate_refresh_token(self, refresh_token_id: str) -> str:
+        refresh_expiration = int((datetime.now(timezone.utc) + timedelta(days=1)).timestamp())
+        refresh_payload = JwtRefreshToken(sub=refresh_token_id, exp=refresh_expiration)
```
IvanObreshkov: The `sub` [JWT claim](https://auth0.com/docs/secure/tokens/json-web-tokens/json-web-token-claims#registered-claims) is used to tell who the subject of this token is, who this token is made for, essentially. There is a separate claim called `jti`, where you can generate an on-demand UUIDv4 to prevent [replay attacks](https://datatracker.ietf.org/doc/html/rfc9700#section-2.2) (for example, [detect that a refresh token has been used more than once](https://auth0.com/blog/refresh-tokens-what-are-they-and-when-to-use-them/#Refresh-Token-Automatic-Reuse-Detection))

See my comments the auth_service.py for more info.

services/py-api/src/service/auth/auth_service.py
```diff
@@ -1,8 +1,165 @@
...
+    ) -> Result[HubAdmin, DuplicateHUBMemberNameError | Exception]:
+
+        # Check if there is another admin with the same name
+        hub_admin_exists = await self._hub_members_repo.check_if_admin_exists_by_name(credentials.name)
+
+        if is_err(hub_admin_exists):
+            return hub_admin_exists
+
+        if hub_admin_exists.ok_value:
+            return Err(DuplicateHUBMemberNameError())
```
IvanObreshkov: This code is redundant. If the user DNE, we will make 2 round-trip requests to the database, which increases latency:

1. We will check whether the user exists
2. If it does, we will make another call to the DB to create it.


Why don't you use the create method directly? If the user DNE, you are going to make 1 roundtrip and create it. If the user exists, the Database Engine will return a DuplikateKeyError which you are going to return directly, which again is one round trip.

services/py-api/src/service/auth/auth_service.py
```diff
@@ -1,8 +1,165 @@
...
+
+        hub_admin = credentials.convert_to_hub_admin()
+        hub_admin.password_hash = self._password_hash_service.hash_password(
+            password_string=credentials.password
+        ).decode("utf-8")
```
IvanObreshkov: Maybe you could pass the hashed_password as a param to the `convert_to_hub_admin` func. In this way, you eliminate the chance of a developer forgetting to assign it and, later, debugging for hours why the password is an empty string.

services/py-api/src/server/handlers/auth/auth_handlers.py
```diff
@@ -1,20 +1,61 @@
...
+    async def register(self, credentials: RegisterHubAdminData) -> Response:
+        result = await self._service.register_admin(credentials=credentials)
+
+        if is_err(result):
+            return self.handle_error(result.err_value)
+
+        return Response(HubMemberSuccessfullyRegistered(hub_admin=result.ok_value), status_code=status.HTTP_201_CREATED)
+
+    async def refresh_token(self, refresh_token: str | None = Cookie(default=None)) -> Response:
+        result = await self._service.refresh_token(refresh_token=refresh_token)
```
IvanObreshkov: The method should be called `refresh_token_pair`

IvanObreshkov (commented):
[Auth Notes.pdf](https://github.com/user-attachments/files/24987032/Auth.Notes.pdf)

A few additional notes:

Bravo! I am proud. This is actually a really good PR, you may feel that it is not, because of the 22 comments, but it is really good. You have followed the patterns we have placed when we created the project, and even created a new one.

1. Could you please update the `is_authorized` route dependency: 

Image

2. I assume you have gotten a bit confused with the whole Access Token and Refresh Token rotation, and the refresh token reuse detection. Yes, it is confusing, I agree :D

You have to read the RFCs and get it wrong a couple of times until you get it right :D, but this is why we are here, and we are a team. You will be surprised, but actually not many devs get how JWTs work, the whole Oauth2.0 flow, and even fewer implement their own Auth from scratch, as you are doing it here.

I am attaching my notes on this topic. I hope you will get the idea from them. I am not saying the code written there is the best one or that this is the best pattern, but I believe you can come up with something even better and more polished than what is done in my document. Good luck!

services/py-api/src/database/model/admin/hub_member_model.py
Bubomira: mongo does not allow HttpUrl to be saved in db, it wants the links as a string

services/py-api/src/service/auth/password_hash_service.py
Bubomira: When i tried to commit my changes, there were error messages saying that the bcrypt methods were actually returning Any, instead of bool/bytes and therefrore i could not return Any from my methods with those signatures.

services/py-api/src/service/auth/auth_service.py
```diff
@@ -1,8 +1,165 @@
...
+
+        if is_err(result):
+            return result
+
+        new_refresh_id = result.ok_value.id
+
+        # Generate new refresh token
+        jwt_refresh_token = self._auth_token_service.generate_refresh_token(refresh_token_id=str(new_refresh_id))
+
+        return Ok((jwt_auth_token, jwt_refresh_token))
```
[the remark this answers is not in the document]
Bubomira: We check whether the refresh token provided is valid(not expired, and token with such id for a hub_member exists in db)
and then delete it when issuing a new one. The decode_refresh_token method would not allow expired tokens, or tokens with invalid schemas. If someone wants to use an old refresh token, it would not be allowed since token with such id would not be available in the db. We save one refresh token in the db per user due to storage issues as well.

PR #1244, opened 2026-02-17
#1089 Add mentors pages in admin panel

### LINKED ISSUE #1089: Implement Admin Panel Mentors page
### Task:
Create the Admin Panel Mentors page for thehub-aubg.com. We are using [shadcn](https://ui.shadcn.com/docs/components) so you can use components from there for easier for implementation.

Page & Routing
-Route: /admin/mentors 
-Route: /admin/mentors/{id} → https://thehub-aubg.com/dashboard/mentors/{id}

-We need to have 3 different pages. 1st page should contain all mentors in card components. You can use the card component from shadcn. 2nd page needs to be a page with input fields which are going to make a request to the backend API to create the mentors. The 3rd page which look like the 2nd page visually but the inputs should be automatically filled with the information of the mentors we are editing

UI Implementation
-Use Ivaka's design from Sponsors/Judges page

### Schemas
Reference Swagger UI at https://localhost:8080/api/v3/docs
They should match case-sensitively

## Wiring task:

Use the generic apiClient.ts at web/src/services and tanstackQuery library to:

Fetch mentors on Mentors page loading
Fetch mentor by id on MentorEditPage  loading
Add, Update mentor on form submit
Delete mentor on delete button press

Note:
You can reference Sponsors Page for example usage
Also with the current sponsors fetching implementation there might be some flicker on page load. You are more than welcome to be creative with any loading animations or anything you come up with.

Schemas:
Available in Swagger UI at: https://localhost:8080/api/v3/docs

#### IMPORTANT: When you test you need the BE running. When you start it for the first time you need to open SwaggerUI and follow the prompt to proceed to site(unsafe) . This is because localhost is not a verified certificate issuer and the browser does not trust it by default.

babajaga3:
> Resolves #1089 

Preview:

image

### DIFF
--- a/services/web/src/types/mentor.ts
+++ b/services/web/src/types/mentor.ts
@@ -7,7 +7,7 @@ const zDateTime = z.string().refine((val) => {
     const zDate = z.string().date()
     const zTime= z.string().time()
 
-    return zDate.parse(date) && zTime.parse(time)
+    return zDate.safeParse(date) && zTime.safeParse(time)
 });
 
 export const zMentor = z.strictObject({


services/web/src/types/mentor.ts
georginikolov7: Why use zod here. Asking of pure curiosity, I feel there is a valid reason I do not see...

PR #1357, opened 2026-08-14
Updated project to use PyMongo async instead of the deprecated Motor dependency

StoyanChorbov:


services/py-api/src/database/repository/hackathon/teams_repository.py
```diff
@@ -82,7 +82,7 @@ async def update(
         self,
         obj_id: str,
         obj_fields: UpdateTeamParams,
-        session: Optional[AsyncIOMotorClientSession] = None,
+        session: Optional[AsyncMongoClient] = None,
```
Bubomira, line 85: ?

services/py-api/src/database/repository/hackathon/teams_repository.py
```diff
@@ -107,7 +107,7 @@ async def update(
             return Err(e)
 
     async def delete(
-        self, obj_id: str, session: Optional[AsyncIOMotorClientSession] = None
+        self, obj_id: str, session: Optional[AsyncMongoClient] = None
```
Bubomira, line 110: here as well

services/py-api/src/service/admin/hub_members_service.py
```diff
@@ -18,7 +20,7 @@ def __init__(self, repo: HubMembersRepository, image_storing_service: ImageStori
     async def get_all(self) -> Result[list[HubMember], Exception]:
         return await self._repo.fetch_all()
 
-    async def get(self, member_id: str) -> Result[HubMember, Exception]:
+    async def get(self, member_id: str) -> Ok[HubMember | HubAdmin] | Err[HubMemberNotFoundError | Exception]:
```
Bubomira, line 23: why have you split the Result like that, out of curiosity?

StoyanChorbov: Because the repo has a more complete exception type and neither PyCharm nor I realized that it should be Result[|] not Ok[] and Err[]

services/py-api/src/database/repository/hackathon/teams_repository.py
StoyanChorbov: Щото блея

